PR #35, opened 2022-05-17
DRAFT: Add vuestic theming

punknoir101:
DRAFT for the vuestic library - for demonstration purposes

Please have a look so that we can decide.

CONS

- small.er community
- docs not very extensive

PROS

- ready to use for VUe3
- a kind of easy to theme via config (variables and on component hierarchy), but with some "downsides" regarding states  (user interactions)
- no storybook addon -->have to write our own


rvp-c:
Only from the point of cleaner (shorter) code I find vuestic better than vuetify.

PR #91, opened 2022-06-22
Feature/e2etest

rvp-c:


rvp-c:
Don't know if we should move the a11y also to the staging verification step

web/playwright.config.ts
```diff
@@ -7,7 +7,7 @@ const config: PlaywrightTestConfig = {
   use: {
     viewport: { width: 1280, height: 720 },
     acceptDownloads: true,
-    baseURL: "http://localhost:4173",
+    baseURL: "https://ris.dev.ds4g.net",
```
zechmeister, line 10: Does this mean, we cannot run e2e tests locally? Wouldn't it be nice, to get the baseUrl from env variables?

web/test/e2e/docunit.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  test("generate doc unit", async ({ page }) => {
+    await page.goto("/")
+
+    const newDocUnitButton = await page.waitForSelector(
+      "button >> text=Neue Dokumentationseinheit"
+    )
+    newDocUnitButton.click()
+
+    await page.waitForSelector("text=Rechtsprechung")
+    docUnitId = await page.locator(".panel-id").innerText()
```
zechmeister, line 15: I think, we should try to avoid using css selectors for tests. Since they rely on implementation details, they tend to be brittle. [Testing-Library](https://testing-library.com/docs/queries/bylabeltext) offers some neat selectors (I'm about to commit the installation).

web/test/e2e/docunit.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+      page.locator("text=Festplatte durchsuchen").click(),
+    ])
+    await fileChooser.setFiles("./test/e2e/sample.docx")
+
+    await page.waitForSelector(
+      ".fileviewer-info-panel-value >> text=sample.docx"
+    )
+  })
+
+  test("delete docx", async ({ page }) => {
```
benjaminaaron, line 37: Just a minor thing: maybe we should call it `delete original file` as we will support more file formats later on?

web/playwright.config.ts
rvp-c: Exactly. The problem is the cloud bucket. Our S3ClientMock doesn't do anything with the uploaded docx file.
The only way for now is to run the backend against staging. 
But yes, we can put the baseUrl into an env variable.

web/test/e2e/docunit.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    const selectDocUnit = await page.waitForSelector(
+      `tr td:nth-child(1) a[href*="/rechtsprechung/${docUnitId}/dokumente"]`
+    )
+    selectDocUnit.click()
+
+    const [fileChooser] = await Promise.all([
+      page.waitForEvent("filechooser"),
+      page.locator("text=Festplatte durchsuchen").click(),
+    ])
+    await fileChooser.setFiles("./test/e2e/sample.docx")
```
benjaminaaron, line 30: Nice that it works that way! I was afraid that part would be more complicated.

web/playwright.config.ts
benjaminaaron: I also think it would be nice to be able to run them locally. It might only work though if we run the profile `staging` locally and not actually the `local` profile? Because we need the real AWS client and not the mockup one 🤔

web/test/e2e/docunit.spec.ts
rvp-c: I'm very interested in this because I know many QA worker, which has this problem with the change of selectors by the developer.

web/test/e2e/docunit.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    await page.goto("/")
+
+    const selectDocUnit = page
+      .locator("tr", {
+        has: page.locator(
+          `td:nth-child(1) a[href*="/rechtsprechung/${docUnitId}/dokumente"]`
+        ),
+      })
+      .locator("td:nth-child(5) i")
+    await selectDocUnit.waitFor()
```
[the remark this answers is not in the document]
rvp-c: it's the same as page.waitForSelector. the problem is the parent/child dependency of the selector. in both cases the result is a promise.

web/playwright.config.ts
carhartl: Also, so far I had tried to keep at least the staging url private, there's a secret for this, which is being populated via environment variable for the journey tests here: https://github.com/digitalservicebund/ris-backend-service/blob/56519668ca152bda8c160abba480f700b14a85b6/src/test/resources/application.properties#L1

rvp-c: done

web/test/e2e/docunit.spec.ts
rvp-c: done

PR #253, opened 2022-11-02
RISDEV-483-Adds frontend routes for norms list and norm view

hamo225:
Adds new base route for /norms

Adds components to display the list of  norms by "longTitle"

Adds route to display a norm by their guid, displaying article title, paragraph text and marker

Adds new menu selection "Normen" in the NavbarTop.

Link to ticket: https://digitalservicebund.atlassian.net/browse/RISDEV-483


web/src/components/NavbarTop.vue
```diff
@@ -25,6 +25,12 @@ const route = useRoute()
         :to="{ name: 'caselaw' }"
         >Rechtsprechung</router-link
       >
+      <router-link
+        class="hover:bg-yellow-500 hover:underline p-8"
+        :class="{ underline: route.path.includes('norms') }"
+        :to="{ name: 'norms' }"
+        >Normen</router-link
+      >
```
weilbith, line 33: Nitpick:
This is a repeating pattern with duplicated code. The purpose of the component is render a list of links. Each of these links should have the same look and effects. At least once a third link gets added, we should refactor this. Till then, it might be fine to just keep it like that. Feel free to decide on your own. 🙃 
With refactoring I mean to have a in the setup function a static list like `[{ label: "Rechtsprechung", routeName: "caselaw" }, {...}]` and then a `v-for` loop in the template. Just as a recommendation. This would also cleanly separate data and logic. 🙃

web/src/components/NormsList.vue
```diff
@@ -0,0 +1,23 @@
...
+            name: 'norms-norm-:normGuid',
+            params: { normGuid: norm.guid },
+          }"
+        >
+          {{ norm.longTitle }}
+        </router-link>
+      </div>
+    </div>
+  </div>
+  <span v-else>Keine Dokumentationseinheiten gefunden</span>
```
weilbith, line 22: Nitpick:
This is now duplicated with the route that does this "error" handling too. Doesn't hurt though.

web/src/components/NormsList.vue
```diff
@@ -0,0 +1,23 @@
+<script lang="ts" setup>
+import Norms from "../domain/Norm"
```
weilbith, line 2: The type is actually called `Norm` in singular form.
I would recommend to remove the default export. So please make `export default Norm` without the `default`. That enforces the usage of the correct name everywhere. No need for a default export anyway. 😊

web/src/domain/Norm.ts
```diff
@@ -0,0 +1,24 @@
...
+  readonly paragraphs: Paragraph[]
+}
+
+export type Paragraph = {
+  guid: string
+  marker: string
+  text: string
+}
+
+export default class Norm {
```
weilbith, line 14: Challenge:
When do you choose to use a type alias and when a class? 🤔

web/src/routes/norms/norm/[:normGuid]/index.vue
```diff
@@ -0,0 +1,29 @@
...
+const norm = ref<Norm | undefined>()
+async function loadNormByGuid() {
+  norm.value = (await getNormByGuid(props.normGuid)).data
+}
+watch(() => props.normGuid, loadNormByGuid, { immediate: true })
+</script>
+<template>
+  <div class="flex flex-col gap-16 p-16">
+    <div v-if="norm" class="p-40 w-[50rem]">
+      <h1 class="heading-02-regular mb-[2.75rem]">
```
Comment (weilbith) [inline web/src/routes/norms/norm/[:normGuid]/index.vue:16]: For the future: please try to stick with the default spacing numbers instead of explicit `rem` values. We try to sync these values properly with design and configure them in the Tailwind setup. Such exceptions make it really hard to achieve a homogeneous look.

web/test/components/normUnitList.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      },
+    })
+
+    await screen.findByText("Keine Dokumentationseinheiten gefunden")
+  })
+
+  test("renders list of norms", async () => {
+    const longTitle = "test"
+    const guid = "123"
+    const articleMock = [
```
weilbith, line 19: No critics at all, rather a knowledge sharing:
We use the concept of data generator functions for domain objects. If you wanna get inspired checkout [this file](https://github.com/digitalservicebund/ris-backend-service/blob/main/web/test/test-helper/dataGenerators.ts). Looking forward to see some norms related generators soon.
For some context: generator functions allow shorter tests that are easier to read and only express the parts that are actually relevant for this specific test case.

weilbith (commented):
Great job. Looks very nice. 💪🏾 
A comment here and there. Mostly just knowledge sharing and small recommendations to make it even better. Decide on your own what do to now or later. 🙂

PR #256, opened 2022-11-04
Implemented Flyway as a database migration system

hmones:
- Added Flyway migrations for both teams: caselaw and norms
- Changed data types of the domain entity for GUIDs to be an instance of UUID
- Created config files for Flyway
- Adjusted tests accordingly to account for changes in the domain data type


build.gradle
```diff
@@ -113,6 +113,10 @@ dependencies {
   // this is to force transitive dependencies to use 1.0.24 instead of
   // 1.0.23 as that is affected by CVE-2022-31684
   implementation "io.projectreactor.netty:reactor-netty-http:1.0.24"
+  runtimeOnly "org.postgresql:postgresql:42.5.0"
+  //  runtimeOnly "org.postgresql:r2dbc-postgresql"
```
weilbith, line 117: Does this comment still serve something? 🤔

src/main/kotlin/de/bund/digitalservice/ris/norms/domain/entity/Norm.kt
```diff
@@ -1,5 +1,5 @@
 package de.bund.digitalservice.ris.norms.domain.entity
 
-import de.bund.digitalservice.ris.norms.domain.value.Guid
+import java.util.UUID
 
-data class Norm(val guid: Guid, val longTitle: String, val articles: List<Article> = listOf())
+data class Norm(val guid: UUID?, val longTitle: String, val articles: List<Article> = listOf())
```
weilbith, line 5: Why are all (`Norm`, `Article` and `Paragraph`) `guid` properties now optional (or rather null-able)? I guess that is a leftover? 🤔 
I mean from the perspective of the Domain, there is never an instance of such an entity without a `guid`. So it should not be nullable, because the whole domain and everyone working with it should not need to handle this null case at all. Because it can't be.

src/main/kotlin/de/bund/digitalservice/ris/norms/framework/adapter/output/NormsMemoryRepository.kt
```diff
@@ -4,21 +4,21 @@ import de.bund.digitalservice.ris.norms.application.port.output.GetAllNormsOutpu
...
+import java.util.*
 
 @Component
 class NormsMemoryRepository : GetAllNormsOutputPort, GetNormByGuidOutputPort, SaveNormOutputPort {
-    private val data = HashMap<Guid, Norm>()
+    private val data = HashMap<UUID, Norm>()
 
     override fun saveNorm(norm: Norm): Mono<Boolean> {
-        data[norm.guid] = norm
+        data[norm.guid]?.let { norm }
```
weilbith, line 17: Se the comment above.

src/main/kotlin/de/bund/digitalservice/ris/norms/framework/adapter/output/database/dto/Article.kt
```diff
@@ -0,0 +1,25 @@
...
+import javax.persistence.CascadeType
+import javax.persistence.FetchType
+import javax.persistence.GeneratedValue
+import javax.persistence.Id
+import javax.persistence.ManyToOne
+import javax.persistence.OneToMany
+import javax.persistence.Table
+
+@Table(name = "articles")
+data class Article(
```
weilbith, line 13: Just as a note. With this class name you'll run into `import ... as...` scenarios where we use a no-mapping strategy for the output ports and provide domain entities as parameter. Those entities than need to be mapped to the data transfer object. So in result you'll need both classes to be available in the same scope, so a name collision will happen. If you prefer alias importing, go ahead. This adapter will become quite independent.

src/main/resources/db/migration/V0.3__norms_create_tables.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE
+  norms (guid uuid PRIMARY KEY, long_title VARCHAR(255));
+
+CREATE TABLE
+  articles (
+    guid uuid PRIMARY KEY,
+    title VARCHAR(255),
+    marker VARCHAR(20),
```
weilbith, line 8: I thought yesterday you said we should be more open about the marker until we understand it better. 😛 
But I guess it is better to start with a small variable character sequence and convert it into a larger one than the other way around. 😆

weilbith (commented):
Nitpick: a commit just focusing on the `Guid` to `UUID` would have been great. Despite other reasons also when we merge this and have to revert it again, then it would be great if the commit would be about one thing. Anyway, no problem to have it like that. 👍🏾 

I'm a bit worried about the *g*uid versus *u*uid thing, but guess there is nothing we can do about that. 

Despite that, see the comment I made. The only really important one for me is the null-able UUID. Please contact me if you disagree.

src/main/kotlin/de/bund/digitalservice/ris/norms/domain/entity/Norm.kt
hmones: The nullable feature of id in specific at domain level is mandatory since we will generate it in the persistence layer and that it should be a unique primary key for sorting in the database. If we are going to use another auto-generated key for indexing I would have agreed with you for not making it nullable. This is the best practice in kotlin especially when you separate domain entities from dtos. You can take a look at this: https://blog.codecentric.de/crud-operations-spring-rest-resources-kotlin

src/main/kotlin/de/bund/digitalservice/ris/norms/framework/adapter/output/NormsMemoryRepository.kt
hmones: Refer to the comment above

src/main/kotlin/de/bund/digitalservice/ris/norms/framework/adapter/output/database/dto/Article.kt
hmones: True, we made use of the alias import option so that we separate the dtos from the entities. This will avoid that the name of the file has something from the package as well. What about if we have this like that and then add it to the confluence page as discussions to settle in our naming conventions? what do you think?

src/main/resources/db/migration/V0.3__norms_create_tables.sql
hmones: I am still open about it as a text, but discussed that with Victor and saw that for the mean time a varchar of 20 is actually a good start. We can increase that later as you said. In the database that shouldn't be a problem as long as we are not changing the type.

build.gradle
hmones: We should remove that, you're right

hmones:
> Please note that work on `main` continued. Means the new input adapters continue to use the `Guid` for now. So this needs to be rebased and adapted. Sorry for that. I guess typical issue of a refactoring PR without a merge/commit freeze.

It would be the same if the checks are failing and I had to push to main :)

src/main/kotlin/de/bund/digitalservice/ris/norms/framework/adapter/output/database/dto/Article.kt
weilbith: Naming convention page/discussion sounds good to me. But in my opinion each adapter can do what ever it wants as long as it is clean code. 🤷🏾

PR #295, opened 2022-12-01
Aktenzeichen: RISDEV-487 + RISDEV-603

benjaminaaron:
Commits on `main` that belong to RISDEV-487 but are not on this PR (oldest first):
- https://github.com/digitalservicebund/ris-backend-service/commit/256050db8363a4b82c584d642fb5dc7afcaa28d6
- https://github.com/digitalservicebund/ris-backend-service/commit/604344f2f2141fd578ffa9b7a5fae6728c32495f
- https://github.com/digitalservicebund/ris-backend-service/commit/6967ab9590a9b3ded81a4daa7b64e3884285efab
- https://github.com/digitalservicebund/ris-backend-service/commit/2f027bc59864c4d88b7dc11a6365936e0bd9618e
- https://github.com/digitalservicebund/ris-backend-service/commit/c81f7bab0b6fb575a2015a6ad22d5621d02a9f60

Commit in the XML exporter repo that belong to RISDEV-487:
- 3656134717ef25145cb15981a60ed0327a5fbfbf


web/src/composables/useTransformTupleData.ts
```diff
@@ -0,0 +1,91 @@
...
+function getTupleKey(parentKey: string, childKey: string) {
+  const sentenceCaseChildKey =
+    childKey.charAt(0).toUpperCase() + childKey.slice(1)
+  return `${parentKey}And${sentenceCaseChildKey}`
+}
+
+function getKeysFromTupleKey(combinedTupleKey: string) {
+  const parentKey = combinedTupleKey.substring(
+    0,
+    combinedTupleKey.indexOf("And")
```
benjaminaaron, line 15: So `And` is used as a string-keyword her to mark this internally. In line 9 it is constructed and here we ask for it 🤔 I wonder if there might be a non-string way to do this? Not sure though. But if it has to be string, a more unique keyword could make sense? I could imagine accidentally an `And` appearing in one of the two keys and then the `indexOf()` would not work properly anymore.

web/src/kitchensink/views/TupleView.vue
```diff
@@ -0,0 +1,32 @@
+<script lang="ts" setup>
+import TupleInput from "@/components/TupleInput.vue"
+import type { InputField } from "@/domain"
+import { defineDateField } from "@/domain"
+
+const tupleFields: { parent: InputField; child: InputField } = {
+  parent: defineDateField(
+    "decisionDate",
+    "Entscheidungsdatum",
```
benjaminaaron, line 9: Why is this a concrete field name here? I would have expected `TupleView.vue` to be abstract since we do the definitions of fields usually in `coreDataFields.ts`?

benjaminaaron (commented):
There are a few blocks of outcommented code. Maybe some of it can be removed? Others make sense to keep like the preparations already for ECLI will be useful later on.

web/src/components/InputElement.vue
```diff
@@ -39,6 +42,10 @@ const component = computed(() => {
       return DateInput
     case InputType.CHECKBOX:
       return CheckboxInput
+    case InputType.MULTITEXT:
```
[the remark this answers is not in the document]
zechmeister: You could easily argue for `chips` instead of Multitext. Although the component only supports `string[]` right now and we'll probably need a dedicated component for multiple dates soon. Right now, the idea is to user MultiText / MultiDate/ Multi... for InputType & Components and extract a `useChips` composable for the multiComponents...

(However, we should probably talk to design and create a shared language)

web/src/composables/useTransformTupleData.ts
zechmeister: I see your point. Since we need IDs for the components, I cannot think of an alternative rn. 

A random `And` should hopefully not brake things, since we only call the function on tuples (see line 47)?

zechmeister: good point! Since, we need the ids I cannot really thing of a good alternative rn. Maybe just check `And` only occurs once?

web/src/kitchensink/views/TupleView.vue
zechmeister: (kitchensink)

web/test/components/multiTextInput.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+    expect(chipList.length).toBe(2)
+    expect(chipList[0]).toHaveTextContent("one")
+    expect(chipList[1]).toHaveTextContent("two")
+
+    await user.type(input, "{arrowleft}")
+    await user.type(input, "{arrowleft}")
+    expect(chipList[0]).toHaveFocus()
+  })
+
+  // it("emits input event when user adds an input", async () => {
```
zechmeister, line 211: I guess we need these tests?

web/src/composables/useTransformTupleData.ts
zechmeister: resolved by [b6d115f](https://github.com/digitalservicebund/ris-backend-service/pull/295/commits/b6d115f69f8d903e8c493540395fa35b3f7582a2)

web/src/components/TupleInput.vue
```diff
@@ -0,0 +1,91 @@
...
+const emit = defineEmits<Emits>()
+
+const { inputValue } = useInputModel<
+  { parent: ModelType; child: ModelType },
+  Props,
+  Emits
+>(props, emit)
+
+const parentValue = computed({
+  get: () => (props.modelValue ? props.modelValue.parent : ""),
```
zechmeister, line 36: I don't think we can just use empty strings here. Tuples should be generic. For example MultiTextInputs expect `string[] | undefined` as `modelValue.` 

(same for line 43)

web/src/components/DocumentUnitCoreData.vue
```diff
@@ -23,11 +24,13 @@ interface Emits {
...
 const emit = defineEmits<Emits>()
+const { modelValue } = toRefs(props)
 
-const values = computed({
-  get: () => props.modelValue ?? {},
-  set: (newValues) => emit("update:modelValue", newValues),
-})
+const values = useTransformTupleData(
+  modelValue,
+  [{ parentKey: "fileNumbers", childKey: "deviatingFileNumbers" }],
```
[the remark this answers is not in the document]
zechmeister: We get the keys from the field definitions now.

web/src/components/DocumentUnitInfoPanel.vue
```diff
@@ -5,11 +5,14 @@ import DocumentUnit from "../domain/documentUnit"
 import PropertyInfo from "@/components/PropertyInfo.vue"
 
 const props = defineProps<{ documentUnit: DocumentUnit }>()
+const fileNumber = props.documentUnit.coreData.fileNumbers
```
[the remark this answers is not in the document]
zechmeister: fixed

web/src/composables/useTransformTupleData.ts
```diff
@@ -0,0 +1,91 @@
...
+      "parent" in tupleValue &&
+      "child" in tupleValue
+    ) {
+      const { parentKey, childKey } = getKeysFromTupleKey(tupleKey)
+
+      const flattenData = { ...newValues }
+      delete flattenData[tupleKey]
+      flattenData[parentKey] = tupleValue.parent
+      flattenData[childKey] = tupleValue.child
+      return flattenData
```
zechmeister: This means, we return work in progress: While we iterate over all the tuples, we return the entire data (still partially tupalized) after every tuple.

web/src/composables/useTransformTupleData.ts
```diff
@@ -0,0 +1,91 @@
...
+    [tupleKey]: {
+      parent: data[parentKey],
+      child: data[childKey],
+    },
+  })
+  return nestedData
+}
+
+function transformToFlatData(newValues: Record<string, ModelType>) {
+  for (const tupleKey in newValues) {
```
zechmeister: This aren't tupleKeys, but all the keys of our data.

web/src/composables/useTransformTupleData.ts
```diff
@@ -0,0 +1,91 @@
...
+    const tupleValue = newValues[tupleKey]
+    if (
+      typeof tupleValue === "object" &&
+      "parent" in tupleValue &&
+      "child" in tupleValue
+    ) {
+      const { parentKey, childKey } = getKeysFromTupleKey(tupleKey)
+
+      const flattenData = { ...newValues }
```
zechmeister: I find verbs for constants mostly unsuitable

web/src/composables/useTransformTupleData.ts
```diff
@@ -0,0 +1,91 @@
...
+      delete flattenData[tupleKey]
+      flattenData[parentKey] = tupleValue.parent
+      flattenData[childKey] = tupleValue.child
+      return flattenData
+      // } else throw new Error(`Can not flatten tuple key: ${tupleKey}`)
```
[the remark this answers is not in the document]
zechmeister: We do not. This iterates over all values. Responsibility to check for valid tuples via regex is moved to `getKeysFromTupleKey`

zechmeister (changes requested):

web/src/composables/useTransformTupleData.ts
zechmeister: [bab5b93](https://github.com/digitalservicebund/ris-backend-service/pull/295/commits/bab5b93392dccfc1067eeff74d6920b1a7db69d5)

web/src/components/MultiTextInput.vue
```diff
@@ -25,9 +25,15 @@ const focusedItemIndex = ref<number>()
 const containerRef = ref<HTMLElement>()
 
 function saveChip() {
-  chips.value.push(currentInput.value)
-  emit("update:modelValue", chips.value.length === 0 ? undefined : chips.value)
-  currentInput.value = ""
+  const trimmed = currentInput.value.trim()
```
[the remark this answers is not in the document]
l30ni3: Added in commit 92e5466

web/src/components/InputField.vue
```diff
@@ -12,6 +12,7 @@ defineProps<Props>()
 <template>
   <div class="flex flex-col flex-start">
     <label
+      v-if="label"
```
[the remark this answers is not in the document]
l30ni3: Added in commit 650d2fb

web/src/components/InputElement.vue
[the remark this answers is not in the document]
zechmeister: decided  for chips (@l30ni3)

web/src/components/DocumentUnitCoreData.vue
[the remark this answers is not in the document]
zechmeister: try to make field names optional and dynamically create tuple field name in coreDataFields.ts (@zechmeister)

web/src/components/MultiTextInput.vue
```diff
@@ -0,0 +1,213 @@
+<script lang="ts" setup>
+import { ref, onMounted } from "vue"
+import { InputField, ValidationError } from "@/domain"
+
+interface Props {
+  id: string
+  value?: string
```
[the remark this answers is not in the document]
zechmeister: @l30ni3

web/src/domain/types.ts
```diff
@@ -16,38 +19,66 @@ export interface BaseInputAttributes {
...
+}
 
+//TUPLE
+export interface TupleInputModelType {
+  parent: ModelType
+  child: ModelType
+}
+
+export interface TupleInputAttributes extends BaseInputAttributes {
+  fields: { parent: InputField; child: InputField }
```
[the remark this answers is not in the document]
zechmeister: @zechmeister tries to use nested objects in both cases

zechmeister (approved):

PR #297, opened 2022-12-02
Some javadoc for some classes

rvp-c:


hmones:
@VictorDelCampo Do you think we still need anything from this branch or should we close it and remove it?

PR #330, opened 2022-12-23
Chore: rename web to frontend for alignment

weilbith:
This is just a PR to trigger the pipeline.


weilbith:
No more necessary. Workflows can be run manually.

PR #334, opened 2022-12-27
Refactor/mono repository structure

weilbith:


hmones:
@weilbith Should we close this and remove the branch or do you think we may still need something out of here?

PR #374, opened 2023-01-17
Error handling

l30ni3:


frontend/test/components/inputElement.spec.ts
```diff
@@ -61,4 +64,23 @@ describe("InputElement", () => {
     expect(emitted()["update:modelValue"]).toHaveLength(1)
     expect(emitted()["update:modelValue"]).toEqual([["a"]])
   })
+
+  it("renders an error message if input is not valid", async () => {
+    renderComponent({
+      id: "Testfeld",
+      type: InputType.DATE,
+      modelValue: "2024-02-10",
```
benjaminaaron, line 72: Maybe +10 years or something? Otherwise next year that test will fail 😉 No biggie though, then we can adjust it of course

frontend/src/components/InputField.vue
```diff
@@ -21,7 +21,7 @@ defineProps<Props>()
       <span v-if="!!required">*</span>
     </label>
 
-    <slot :id="id" />
+    <slot :id="id"> </slot>
```
benjaminaaron, line 24: Was that a suggestion from your IDE? That space in between though should probably go 🤔

frontend/src/components/InputElement.vue
```diff
@@ -69,6 +80,9 @@ const value = computed({
     :id="id"
     v-model="value"
     v-bind="attributes"
-    :validation-error="validationError"
+    v-model:validation-error="validationError"
   />
+  <div class="h-16 label-03-reg text-red-800">
```
benjaminaaron, line 85: Maybe we can add a comment here that this is the temporary location and should be refactored once we talked about error handling soon? Because now there is two locations, which might be confusing to developers coming across this.

benjaminaaron (approved):

frontend/src/components/InputField.vue
l30ni3: yes! but i put the shorthand back in :)

frontend/src/components/InputElement.vue
l30ni3: done

PR #485, opened 2023-04-05
Avoid duplicate ids

l30ni3:
We make the assumption, that the field id in the frontend always matches the domain model in order to communicate with the backend. This resulted in a duplicate-id-aria violation, as domain names appear more often on a page (e.g. court). In order to enable the rule again, an extra “id” property is added to the field configs, to decouple id from domain name if needed, so we are able to use different ids in the frontend.


weilbith:
@l30ni3 is this PR still relevant and should stay open?

PR #676, opened 2023-07-31
Refactor: rework chips components

andreasphil:
This affects:

- Chips list
- Chips input
- Keywords chips

These use shared logic and very similar in terms of implementation and interaction (even though there are some layout and requirement details that make it more convenient to keep them separate). Note that the date chips input is not affected, since it's completely separate from the others.

This commit:

- Changes the list and chips v-models to be immutable according to the new immutable v-model pattern (see ADR 15)
- Also refactors the focus handling to be a v-model to make it easier to use and less verbose
- Fixes a bunch of bugs in the chips and components using the chips along the way
- Improves test coverage for all 3 components, splits the tests for keywords and regular chips into 2 files

RISDEV-2038

### DIFF
--- a/frontend/src/components/KeyWords.vue
+++ b/frontend/src/components/KeyWords.vue
@@ -71,8 +71,8 @@ watch(
           aria-label="Schlagwörter"
           :error="errorMessage"
           :model-value="keywords"
-          @add-chip="addKeyword"
-          @delete-chip="deleteKeyword"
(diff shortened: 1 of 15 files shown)


l30ni3 (approved):

andreasphil:
Has been merged but GitHub somehow missed it, closing the PR

PR #819, opened 2023-10-24
Use LocalDate, change date format

leonie-koch:
Right now the publishing is not yet working. In order to work properly, the exporter needs to be updated. Right now it fails [here](https://github.com/digitalservicebund/neuris-juris-xml-export/blob/1753006aa7720a02d0b11d4e34fef2291463c7f8/src/main/java/de/bund/digitalservice/ris/domain/export/juris/JurisDocumentBuilder.java#L92), where it expects an ISO formatted string with a time value.


weilbith (commented):
After a quick look, I think that should be fine. Anything you are worried about?

PR #883, opened 2023-11-21
remove old env.slack_webhook_url check

patjouk:
Right now, the notify step will always be skipped since it's checking for an env (`env.SLACK_WEBHOOK_URL`) that is not set anymore. Example: https://github.com/digitalservicebund/ris-backend-service/actions/runs/6945318556/job/18894392392


HPrinz:
Thank you!

PR #1178, opened 2024-02-05
feat: Add help page to search

madebyherzblut:
Fixes RISDEV-3217, RISDEV-3216


benjaminaaron (approved):
Looks great! Very nice with the outsourcing to `data.ts`. Maybe we can actually reuse the type definitions in there for the search operations - in case we need to be more specific than string there.

PR #1349, opened 2024-04-08
Multiple singlenorms

leonie-koch:
Changelog:

- Update the UI according to designs
- Wire up the grouped norm components with the backend logic and the data flow
- Abbreviation update does not trigger emit -> effect is, that the cancel and delete button not visible in initial 'creation' mode
- Updating the norms summary logic
- Fixing the display of norms summary, with no/ empty single norms
- Adding same abbreviation twice not possible

Design: https://www.figma.com/file/oyeWgl6BRJA15tkGjUD6Ze/neuRIS-Rechtsprechung?type=design&node-id=12695-17801&mode=design&t=ahymZby7k8JEMl1n-0

Ticket: https://digitalservicebund.atlassian.net/jira/software/c/projects/RISDEV/boards/7?assignee=62bad547ed036549273cd7de&selectedIssue=RISDEV-3546


frontend/src/components/NormReferences.vue
```diff
@@ -23,23 +24,40 @@ const norms = computed({
   },
 })
 
+function hasSingleNorms(normEntry: NormReference) {
+  if (normEntry.singleNorms)
+    return (
+      normEntry.singleNorms?.length > 1 || !normEntry.singleNorms[0].isEmpty
```
anne-ds, line 30: When length is 0 this line will throw an TypeError: Cannot read properties of undefined (reading 'isEmpty').
Use case: Use deletes all singleNorms.

leonie-koch: Gut gesehen! hab ich korrigiert

rvp-c: But is normEntity.singleNorms not false if it's a empty array?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/NormReferenceDTO.java
```diff
@@ -31,6 +31,9 @@ public class NormReferenceDTO {
   @JoinColumn(name = "norm_abbreviation_id")
   private NormAbbreviationDTO normAbbreviation;
 
+  @Column(name = "norm_abbreviation_raw_value", insertable = false, updatable = false)
+  private String normAbbreviationRawValue;
```
anne-ds, line 35: Is the introduction of the rawValue aligned with the YAGNI-Principle? The frontend is not using it at this point, right?

rvp-c: But I hope we do it with this refactoring too?

frontend/src/components/NormReferenceInput.vue
```diff
@@ -87,23 +100,15 @@ watch(
     norm.value = new NormReference({ ...props.modelValue })
     lastSavedModelValue.value = new NormReference({ ...props.modelValue })
     if (lastSavedModelValue.value.isEmpty) validationStore.reset()
+    //when list is empty, add new emptry single norm entry
```
anne-ds, line 103:
```suggestion
    //when list is empty, add new empty single norm entry
```

frontend/src/components/NormReferenceInput.vue
```diff
@@ -121,97 +126,56 @@ onBeforeUnmount(() => {
...
-            v-if="!lastSavedModelValue.isEmpty"
-            aria-label="Abbrechen"
-            button-type="ghost"
-            label="Abbrechen"
-            size="small"
-            @click.stop="emit('cancelEdit')"
-          />
+    <div v-if="normAbbreviation">
+      <SingleNormInput
+        v-for="(entry, index) in singleNorms"
```
anne-ds, line 135:
```suggestion
        v-for="(_, index) in singleNorms"
```

frontend/src/components/NormReferences.vue
anne-ds: Hm, I don't see your fix 🤔 ? Did everything work fine git-wise?

leonie-koch: This condition seems to work though :)

PR #1398, opened 2024-04-18
Add ErrorBadge for ambiguous norm reference (RISDEV-3700)

anne-ds:
RISDEV-3700


leonie-koch (commented):
This is great implemented! Nothing to complain :)

PR #1542, opened 2024-05-28
Show only procedures with docUnits

anne-ds:
RISDEV-3795

I had to refactor the ProcedureIntegrationTest as there has been a mix in the data setup and also duplicates in the tests. With this MR the tests have a streamlined data setup in a common init script.


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/DatabaseProcedureRepository.java
```diff
@@ -12,18 +12,22 @@
 @Repository
 public interface DatabaseProcedureRepository extends JpaRepository<ProcedureDTO, UUID> {
```
leonie-koch, line 13: do we want java doc for this repository interface?

anne-ds: Good idea. Thanks. Done ✅

PR #1583, opened 2024-06-03
Improve invisible-characters

anne-ds:
RISDEV-4026


frontend/src/components/input/TextEditor.vue
```diff
@@ -318,12 +318,26 @@ watch(
...
+watch(
+  () => hasFocus.value,
+  () => {
+    if (props.editable) {
+      // When the TextEditor is editable and has focus, the invisibleCharacters should be displayed
+      commands.toggleActiveState()(editor.state, editor.view.dispatch)
+    }
+  },
+)
+
```
leonie-koch, line 330: I would remove the commands.setActiveState from the onMounted method and put it all in one place:

```suggestion
watch(
  () => hasFocus.value,
  () => {
    if (props.editable) {
      // When the TextEditor is editable and has focus, the invisibleCharacters should be displayed
      commands.setActiveState(props.editable && hasFocus.value)(
        editor.state,
        editor.view.dispatch,
      )
    }
  },
  { immediate: true },
)

```

anne-ds: Thanks for the hint 🙏 . I changed it and love the simplified version. However, I added it without the if-statement as it is not needed.

PR #1614, opened 2024-06-06
Refactoring field of law

Nadav-B:
RISDEV-4030

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/transformer/FieldOfLawTransformer.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/transformer/FieldOfLawTransformer.java
@@ -43,6 +43,10 @@ public class FieldOfLawTransformer {
       builder.linkedFields(linkedFields);
     }
 
+    if (fieldOfLawDTO.getChildren() != null) {
+      builder.hasChildren(!fieldOfLawDTO.getChildren().isEmpty());
+    }
+
     if (withChildren && fieldOfLawDTO.getChildren() != null) {
       List<FieldOfLaw> children =
           fieldOfLawDTO.getChildren().stream()
@@ -50,15 +54,10 @@ public class FieldOfLawTransformer {
               .toList();
       if (!children.isEmpty()) {
         builder.children(children);
-        builder.childrenCount(children.size());
+        builder.hasChildren(true);
       }
     } else {
       builder.children(Collections.emptyList());
-      builder.childrenCount(-1);
-      // TODO
-      //      if (fieldOfLawDTO.getChildren() != null) {
-      //        builder.childrenCount(fieldOfLawDTO.getChildren().size());
-      //      }
     }
 
     return builder.build();
--- a/frontend/src/components/FieldOfLawMain.vue
+++ b/frontend/src/components/FieldOfLawMain.vue
@@ -6,17 +6,17 @@ import { withSummarizer } from "@/components/DataSetSummary.vue"
 import ExpandableDataSet from "@/components/ExpandableDataSet.vue"
 import FieldOfLawDirectInputSearch from "@/components/FieldOfLawDirectInputSearch.vue"
 import FieldOfLawSearch from "@/components/FieldOfLawSearch.vue"
-import { FieldOfLawNode } from "@/domain/fieldOfLaw"
+import { FieldOfLaw } from "@/domain/fieldOfLaw"
 
 const props = defineProps<{
-  modelValue: FieldOfLawNode[] | undefined
+  modelValue: FieldOfLaw[] | undefined
 }>()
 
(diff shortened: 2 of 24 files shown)


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/transformer/FieldOfLawTransformer.java
```diff
@@ -43,22 +43,21 @@ public static FieldOfLaw transformToDomain(
       builder.linkedFields(linkedFields);
     }
 
+    if (fieldOfLawDTO.getChildren() != null) {
+      builder.hasChildren(!fieldOfLawDTO.getChildren().isEmpty());
```
leonie-koch, line 47: [Non-blocking]
Can't we just say builder.hasChildren(true); like in line 57? is there a difference to set it differently?

frontend/src/components/FieldOfLawMain.vue
```diff
@@ -30,7 +30,7 @@ watch(localModelValue, () => {
   emit("update:modelValue", localModelValue.value)
 })
```
leonie-koch, line 32: [Non-blocking]
I actually thought of removing all the logic of adding/removing nodes and also the "selectedNode" from the FieldOfLawMain component and just  have the modelValue here as single source of truth. When a search result is clicked in FieldOfLawSearch, the component emits an event and the modelValue ist updated. When a tree node is clicked in FieldOfLawTree, the same happens. And when a result in FieldOfLawDirectInputSearch was clicked, again the same. So the Main component would hold the modelValue, and all three components update this modelValue. We wouldn't need all these functions (setSelectedNode, removeSelectedNode, addFieldOfLaw, removeFieldOfLaw)in the main component for example and for each component just one emit (@update:modelValue) (instead all those other listeners). 

But as this a big refactoring and also out of scope of this Mangel ticket, this is just a nice to have comment. Also, I possible oversee things, which make a refactoring like this not possible.

frontend/src/components/FieldOfLawNodeComponent.vue
```diff
@@ -1,178 +1,186 @@
...
+  selectedNode?: FieldOfLaw
 }
 
 const props = defineProps<Props>()
+
 const emit = defineEmits<{
-  "node:select": [node: FieldOfLawNode]
-  "node:unselect": [node: FieldOfLawNode]
-  "linkedField:clicked": [identifier: string]
+  "node:select": [node: FieldOfLaw]
```
leonie-koch, line 24: [Non-blocking]
This again just wishful thinking of a bigger refactoring of this component: 
Can't we just emit a `selected` boolean value? How does the selected value differ from linkedField:clicked? Do we actually need to emit expand and collapse, because we can derive the information also from the `selected` value?

frontend/src/components/FieldOfLawMain.vue
Nadav-B: @leonie-koch if it makes sense, we can switch to v-model. However, it is still does not change the matter that we currently based on the design we always need: 
selected node -> The node that is currently shown on the tree
saved nodes (modelValue for instance) -> the field of law nodes that is saved in our backend. 
Based on the design we need to have both always, to once have a direct tree to the node, second: show the user if this node is already saved in document unit

frontend/src/components/FieldOfLawNodeComponent.vue
Nadav-B: One needs to know the selected node. The selected node is that object that came from either from the search or from the other list. The tree needs to know which node to show the direct path to. The boolean value is not enough. `linkedField` is a mystery for me. I don't know why, anyway, we had a linked field that is differed from node fieldOfLaw. In my offer I created a node out of this linkedField but did not want to get rid of it completely as now I know clearer and clearer that is just not needed in code.

frontend/src/components/fieldOfLawNode.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+      } catch (error) {
+        return []
+      }
+    } else {
+      return []
+    }
+  }
+
+  async getFilteredChildren(
```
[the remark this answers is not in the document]
Nadav-B: In this way we can remove getFilteredChildren and directly call the getChildren with the node that we need. I would also like to get rid of the nested nodes that the backend is giving as it just make us a harder work to extract them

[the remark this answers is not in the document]
leonie-koch: Yes, let's keep that in mind for the next refactoring, when there is als UX imrovements, we do it all together :)

frontend/src/components/FieldOfLawMain.vue
leonie-koch: 💡

frontend/src/components/FieldOfLawNodeComponent.vue
leonie-koch: 💡Ok good to know!

PR #1619, opened 2024-06-07
RISDEV-3503: Umgebungen optisch unterscheidbar machen

elaydis:


backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/CurrentEnvironment.java
```diff
@@ -0,0 +1,3 @@
+package de.bund.digitalservice.ris.caselaw.domain;
+
+public record CurrentEnvironment(String name) {}
```
leonie-koch, line 3: I am not sure if I fully understood where the enviornment comes from - but does it make sense to give CurrentEnvirionment an enum ("production" | "staging" | "uat") ?

frontend/vite.config.ts
```diff
@@ -12,6 +12,11 @@ export default defineConfig({
   build: {
     sourcemap: true, // Source map generation must be turned on
   },
+  esbuild: {
+    supported: {
+      "top-level-await": true,
```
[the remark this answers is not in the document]
elaydis: Yes, exactly. I guess we could also move the session store initialisation, but it seems to me that this should not be an issue based on https://caniuse.com/mdn-javascript_operators_await_top_level stats. I only found this problem because the pipeline failed [here](https://caniuse.com/mdn-javascript_operators_await_top_level).

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/CurrentEnvironment.java
elaydis: It comes from the spring boot config yaml files, so I felt like using an enum would make this more cumbersome and not actually safer, but I am happy to be challenged here :) basically now in `EnvironmentConfig` line 11 it gets read directly from the yaml file as a string. I guess we could cast it to an enum there, but not sure if it brings a huge benefit 🤔  what do you think?

elaydis:
@leonie-koch don't be surprised. I am already merging the PR to test that publishing still works, but let's continue talking about your suggestions :)

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/CurrentEnvironment.java
leonie-koch: Ok great, got it, makes sense to just take the string here :)

PR #1651, opened 2024-06-12
Refactor combobox options 

kiwikern:
RISDEV-4079


frontend/src/components/input/types.ts
```diff
@@ -136,6 +136,7 @@ export type ComboboxInputModelType =
 export type ComboboxItem = {
   label: string
   value?: ComboboxInputModelType
+  labelCssClasses?: string
```
leonie-koch, line 139: Super good! I was wondering if we find a better way to distinguish between "normal" items and the "createNewItem", because this is the only case, where we would give special css. Does make sense to add a property to the combobox item?  Or can't we just check in the :class if the label equals the `${searchStr} neu erstellen`

leonie-koch: But the more I think about it ... checking for label === `${searchStr} neu erstellen` is also not a nice option. We keep it like this! :)

leonie-koch (commented):
Can be merged 👍

frontend/src/components/input/types.ts
kiwikern: Thank you for your feedback! 🙏 Yes, I also thought about it and couldn't really decide which way is better. I know opted for the "more general" solution, but not sure if it is the better way.
Then I will merge this now, but also happy to revisit this. :-)

PR #1652, opened 2024-06-12
Copy and paste from editable list summary

leonie-koch:
RISDEV-4028

### DIFF
--- a/frontend/src/components/ActiveCitationSummary.vue
+++ b/frontend/src/components/ActiveCitationSummary.vue
@@ -1,5 +1,6 @@
 <script lang="ts" setup>
 import { computed } from "vue"
+import DecisionSummary from "@/components/DecisionSummary.vue"
 import IconBadge from "@/components/IconBadge.vue"
 import ActiveCitation from "@/domain/activeCitation"
 import IconBaselineContentCopy from "~icons/ic/baseline-content-copy"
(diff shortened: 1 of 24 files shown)


frontend/src/components/ActiveCitationSummary.vue
Nadav-B, line 1: I love it, great job exporting this summary component out, reducing the code lines by reusing already existing and tested components.

frontend/src/domain/relatedDocumentation.ts
Nadav-B, line 1: Happy to see this part out. A long string in favor of our variables limits our flexibility to customize styling to part of it.

frontend/src/components/ActiveCitationSummary.vue
Nadav-B: @leonie-koch question here: in our [code](https://github.com/digitalservicebund/ris-backend-service/commit/fdfa79707073a533a0f53bcff6607a7ca4e948bd#diff-76c4fc531deecc6de1fdbec98285ee7eaff245595026459cca75cbb7dd5f8c11) we once use test id in camel and once in underscore. 

`await page.getByTestId("listEntry-0").click()`

[Can we decide for the official guideline of the library](https://testing-library.com/docs/queries/bytestid/#api)?

leonie-koch: Absolutely! Haven't seen that, thanks for pointing out, i will change that 👍

PR #1701, opened 2024-06-19
Display note

kiwikern:


frontend/src/components/DocumentUnitAttachments.vue
```diff
@@ -125,20 +104,10 @@ async function upload(files: FileList) {
...
 
-const toggleExtraContentPanel = (state?: boolean) => {
-  showExtraContentPanelRef.value = state || !showExtraContentPanelRef.value
-  pushQueryToRoute({
-    showAttachmentPanel: showExtraContentPanelRef.value.toString(),
-  })
-}
-
 function toggleDeleteModal() {
   showDeleteModal.value = !showDeleteModal.value
```
kiwikern, line 112: We should move the checks whether the attachment index actually exists from the template to here. Otherwise, it might not be possible to close the dialog. Also, I think I prefer a open and close method instead of toggle, because the user operations are distinct. (close dialog button vs. delete file icon)

frontend/src/components/ExtraContentSidePanel.vue
```diff
@@ -100,41 +116,50 @@ watch(
...
-          ></FileNavigator>
           <AttachmentView
             v-if="
               documentUnit.uuid &&
               documentUnit.attachments &&
-              props.currentIndex != null &&
-              documentUnit.attachments[props.currentIndex] &&
-              documentUnit.attachments[props.currentIndex]?.s3path
+              currentAttachmentIndex != null &&
+              documentUnit.attachments[currentAttachmentIndex] &&
```
kiwikern, line 154: can be removed

frontend/src/components/DocumentUnitAttachments.vue
elaydis: very good point, let's fix it :)

PR #1736, opened 2024-06-25
RISDEV-4117 - Add indentation feature to editor

HPrinz:
A paragraph can be indented with buttons or Tab key. To unindent, another button or Shift + Tab key can be used. The maximal level of indentations is 10

### DIFF
--- a/frontend/src/editor/indent.ts
+++ b/frontend/src/editor/indent.ts
@@ -1,0 +1,215 @@
+/* eslint-disable no-param-reassign */
+// Sources:
+// https://github.com/ueberdosis/tiptap/issues/1036#issuecomment-981094752
+// https://github.com/django-tiptap/django_tiptap/blob/main/django_tiptap/templates/forms/tiptap_textarea.html#L453-L602
+
+import {
+  CommandProps,
+  Extension,
+  Extensions,
+  isList,
+  KeyboardShortcutCommand,
+} from "@tiptap/core"
+import { TextSelection, Transaction } from "prosemirror-state"
+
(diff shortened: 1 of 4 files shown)


frontend/src/editor/indent.ts
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable no-param-reassign */
```
anne-ds, line 1: Do we need this? Why?

frontend/src/editor/indent.ts
```diff
@@ -0,0 +1,189 @@
...
+  HTMLAttributes: Record<string, unknown>
+}
+
+type IndentType = "indent" | "outdent"
+
+const clamp = (val: number, min: number, max: number): number =>
+  Math.min(Math.max(val, min), max)
+
+const updateIndentLevel = (
+  tr: Transaction,
```
anne-ds, line 40: I personally often have difficulties with abbreviations in code. At first I thought: "What does table row have to do with this?". What do you think about renaming this to "transaction". I think it is not too long.

frontend/src/editor/indent.ts
```diff
@@ -0,0 +1,189 @@
...
+        },
+    }
+  },
+
+  addKeyboardShortcuts() {
+    return {
+      Tab: getIndent(),
+      "Shift-Tab": getOutdent(false),
+      Backspace: getOutdent(true),
+      "Mod-]": getIndent(),
```
anne-ds, line 175: I thought we have just the tab command... What kind of command is this? How can I trigger it?

PR #1746, opened 2024-06-26
RISDEV-4120 - Convert indentation from docx cocuments to HTML indentation (margin-left style)

HPrinz:


anne-ds:
Looks good to me 👍

PR #1928, opened 2024-07-18
Fix xml trivy warning

kiwikern:


kiwikern:
@HPrinz the [`ConverterConfig`](https://github.com/digitalservicebund/ris-backend-service/blob/83fd046fca64c7998c73dc8354dae765a344cc06/backend/src/main/java/de/bund/digitalservice/ris/caselaw/config/ConverterConfig.java#L26-L26) already has a `DocumentBuilderFactory` that conforms with the requirements. Not sure if we can unify these!?

PR #2065, opened 2024-08-13
Make core data read-only for externals

anne-ds:
RISDEV-4523


frontend/test/e2e/caselaw/categories/coreData/core-data.spec.ts
```diff
@@ -307,4 +307,194 @@ test.describe("core data", () => {
...
     ).toBeHidden()
   })
+
+  test(
+    "core data is readonly for external user",
+    {
+      annotation: {
+        type: "story",
+        description:
+          "https://digitalservicebund.atlassian.net/browse/RISDEV-4523",
```
kiwikern, line 317: non-blocking: Nadav found out that you can add stories/epics also as `tag` which will be searchable in the test report, see https://github.com/digitalservicebund/ris-backend-service/blob/abfccc81a88e7f8e815b596b3b84871054ed5f61/frontend/test/e2e/caselaw/categories/shortAndLongTexts/short-and-long-texts.spec.ts#L61-L61 Then it's easier to share for the testers. But we did not commit to doing this always, so no requirement.

kiwikern (approved):
LGTM! Great job! 👏 🚀 🎉

PR #2135, opened 2024-09-06
Edition handover section 

HPrinz:
RISDEV-4745


backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/HandoverService.java
```diff
@@ -85,19 +89,37 @@ public XmlTransformationResult createPreviewXml(UUID documentUuid)
...
    * Get the event log for a documentation unit, containing jDV email handover operations, handover
    * reports (response emails from the jDV) and migrations/import events.
    *
-   * @param documentUuid the UUID of the documentation unit
+   * @param entityId the UUID of the documentation unit
    * @return the event log
    */
-  public List<EventRecord> getEventLog(UUID documentUuid) {
+  public List<EventRecord> getEventLog(UUID entityId, HandoverEntityType entityType) {
+    if (entityType == HandoverEntityType.EDITION) {
```
Nadav-B, line 114: For clarity, it seems like we can use switch cases:

case (DOCUMENTATION_UNIT)...
case(Edition)....

I usually not a fan of switch for two cases only but comined with enum it makes sense

backend/src/main/resources/db-scripts/migration/V1.8__allow_multiple_attachments_on_handover.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE IF NOT EXISTS handover_mail_attachments (
```
Nadav-B, line 1: Question of interest: When do we keep the attachment in the database? Is it only when an attachment was successfully sent?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/LegalPeriodicalEditionDTO.java
```diff
@@ -31,7 +30,7 @@
 @Getter
 @Setter
 public class LegalPeriodicalEditionDTO {
-  @Id @GeneratedValue private UUID id;
+  @Id private UUID id;
```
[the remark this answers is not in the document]
HPrinz: I tried with overriding the Id, but I ran into problems there that I don't remember exactly anymore. We could generate the Id in the transformer if it's not set already -WDYT?

backend/src/main/resources/db-scripts/migration/V1.8__allow_multiple_attachments_on_handover.sql
HPrinz: Yes, as soon as we sent the mail

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/HandoverService.java
HPrinz: Good idea! I changed it that way!

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/HandoverRepository.java
```diff
@@ -6,7 +6,7 @@
 public interface HandoverRepository {
   HandoverMail save(HandoverMail handoverMail);
 
-  List<HandoverMail> getHandoversByDocumentationUnitId(UUID documentationUnitId);
+  List<HandoverMail> getHandoversByEntity(UUID documentationUnitId, HandoverEntityType entityType);
```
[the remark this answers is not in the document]
HPrinz: Yes, actually there were similar mistakes in other files that I now corrected 😇

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/LegalPeriodicalEditionDTO.java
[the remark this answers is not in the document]
Nadav-B: Currently with the other editable list items for instance, we generate an id at frontend side and when saving the backend overrides it and we fetch the new id. 

We either can leave it for now and say: 
1. frontend sends the id upfront, when a new legal periodical edition is created
 or we align with the other id creation policy, meaning:
2. we make a save call on before mounted to the backend, fetch the created id and push it to the route parameters. We this behaviour we probably do the same already with document unit 

In this case the client does not have any control over the actual generated id.

PR #2187, opened 2024-09-24
New list input component

leonie-koch:


frontend/src/components/input/ListInput.vue
```diff
@@ -0,0 +1,145 @@
...
+
+onMounted(() => {
+  // When data, show display mode. When empty, show edit mode
+  editMode.value = !props.modelValue.length
+})
+</script>
+
+<template>
+  <div>
+    <!-- Edit mode -->
```
kiwikern, line 72: suggestion: I would split this component into edit and display component

frontend/src/components/input/ListInput.vue
```diff
@@ -0,0 +1,145 @@
...
+
+// Watch the `textAreaInput` value to adjust the height when content changes
+watch(textAreaInput, async () => {
+  await nextTick() // Wait for DOM render
+  const textarea = document.querySelector("textarea")
+  adjustTextareaHeight(textarea as HTMLTextAreaElement)
+})
+
+onMounted(() => {
+  // When data, show display mode. When empty, show edit mode
```
kiwikern, line 65: This comment describes what the code below does. I would prefer the "why" over the "what", e.g.: Without a value, the component is collapsed by the CategoryWrapper. When expanding it, it should be in edit mode right away.

frontend/src/components/input/ListInput.vue
```diff
@@ -0,0 +1,145 @@
...
+      </div>
+    </div>
+    <!-- Display mode -->
+    <div v-else class="flex flex-col gap-16">
+      <div class="flex flex-col gap-8">
+        <span class="ds-label-02-reg mb-4">{{ label }}</span>
+        <ul class="m-0 flex flex-row gap-8 p-0">
+          <li
+            v-for="(chip, i) in modelValue"
+            :key="i"
```
kiwikern, line 124: At least in React, it is better not to use an index for the key, but rather a unique value that identifies the entity. For us that would just be the string as it is unique. The framework will use the key to determine whether a list element needs to be re-rendered or can be reused. When the order changes (so also when elements are removed or added), the same index might refer to a different "entity".
https://react.dev/learn/rendering-lists#why-does-react-need-keys

frontend/src/components/input/ListInput.vue
```diff
@@ -0,0 +1,145 @@
...
+
+const emit = defineEmits<{
+  "update:modelValue": [value: string[]]
+}>()
+
+const editMode = ref(true)
+const sortAlphabetically = ref(false)
+const localModelValue = ref(props.modelValue ?? [])
+
+const textAreaInput = computed({
```
kiwikern, line 20: If we split the EditComponent into its own value, I would rather pass the initial value as a prop and then keep the state locally in the component and emit it once the submit button is clicked. Would be happy to discuss this, as the consequences are not clear to me.
While trying the feature, I noticed that the server values override the local values. (But that might be a local problem.)

kiwikern (commented):
It looks really nice and clean, amazing how quickly you got this together. 👏 🥳 
I left a few comments, but couldn't finish the review as I have a lunch date now. Will continue later.

frontend/src/components/input/ListInput.vue
```diff
@@ -0,0 +1,145 @@
...
+    localModelValue.value = localModelValue.value.sort((a: string, b: string) =>
+      a.localeCompare(b),
+    )
+  }
+  emit("update:modelValue", [...new Set(localModelValue.value)] as string[]) //remove duplicates
+  if (!!localModelValue.value?.length) editMode.value = false
+  sortAlphabetically.value = false
+}
+const cancelEdit = () => {
+  if (!!localModelValue.value?.length) editMode.value = false
```
kiwikern, line 43: For me as a user, it was confusing that I cannot leave the edit mode when there is no input. I thought it was not saving it. I asked Agnieszka [here](https://www.figma.com/design/oyeWgl6BRJA15tkGjUD6Ze?node-id=18303-63930#962540636).

frontend/src/components/input/ListInput.vue
```diff
@@ -0,0 +1,145 @@
...
+            @click.stop="cancelEdit"
+          />
+        </div>
+      </div>
+    </div>
+    <!-- Display mode -->
+    <div v-else class="flex flex-col gap-16">
+      <div class="flex flex-col gap-8">
+        <span class="ds-label-02-reg mb-4">{{ label }}</span>
+        <ul class="m-0 flex flex-row gap-8 p-0">
```
kiwikern, line 121: I would add a `flex-wrap` here in case the list overflows.

frontend/src/components/input/ListInput.vue
```diff
@@ -0,0 +1,145 @@
...
+  // When data, show display mode. When empty, show edit mode
+  editMode.value = !props.modelValue.length
+})
+</script>
+
+<template>
+  <div>
+    <!-- Edit mode -->
+    <div v-if="editMode" class="flex flex-col gap-24">
+      <div class="flex flex-col gap-8">
```
kiwikern, line 74:
```suggestion
      <div class="flex flex-col">
```
In the design we only have a gap of 4px which the label already has as `mb`.
image

frontend/src/components/input/ListInput.vue
kiwikern, line 121:
```suggestion
        <ul class="m-0 flex flex-row p-0">
```
Same as in edit mode, the design has only 4px from the label's `mb`

frontend/src/components/input/ListInput.vue
kiwikern: It seems that, as opposed to React, for vue, the `key` attribute is actually not needed for simple list entities
> When Vue is updating a list of elements rendered with v-for, by default it uses an "in-place patch" strategy. If the order of the data items has changed, instead of moving the DOM elements to match the order of the items, Vue will patch each element in-place and make sure it reflects what should be rendered at that particular index.
> This default mode is efficient, but only suitable when your list render output does not rely on child component state or temporary DOM state (e.g. form input values).

https://vuejs.org/guide/essentials/list#maintaining-state-with-key

frontend/src/components/input/ListInput.vue
```diff
@@ -0,0 +1,145 @@
...
+        <label id="list-input" class="ds-label-02-reg mb-4"> {{ label }}</label>
```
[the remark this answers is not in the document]
kiwikern: We keep it.

frontend/src/components/input/ListInput.vue
kiwikern: Agnieszka wants the state to go back to the CategoryWrapper button. Maybe we can add an optional reset handler to the CategoryWrapper that we pass to the slot, see https://vuejs.org/guide/components/slots.html#scoped-slots

frontend/src/components/input/ListInput.vue
leonie-koch: Great point. Done here:
https://github.com/digitalservicebund/ris-backend-service/pull/2187/commits/fc84e057b5e083c641cfba28f56440e483c11be2

https://github.com/digitalservicebund/ris-backend-service/pull/2187/commits/4cd7a8115bf043d8599352c2b154ae42734bb63d

frontend/src/components/input/ListInput.vue
leonie-koch: Great point. Done here:
https://github.com/digitalservicebund/ris-backend-service/pull/2187/commits/fc84e057b5e083c641cfba28f56440e483c11be2

https://github.com/digitalservicebund/ris-backend-service/pull/2187/commits/4cd7a8115bf043d8599352c2b154ae42734bb63d

frontend/src/components/input/ListInput.vue
leonie-koch: Good catch, should be fixed with the edit component now :)

frontend/src/components/input/ListInput.vue
leonie-koch: Thank you for evaluating with Agnieszka, added it here :)
https://github.com/digitalservicebund/ris-backend-service/pull/2187/commits/8527415da81ffc265afa471593395bd37ae083c1

frontend/src/components/input/ListInput.vue
leonie-koch: Great catch: https://github.com/digitalservicebund/ris-backend-service/pull/2187/commits/d861e6cb1eba6811405e493b671774bd2e68a9f7

PR #2313, opened 2024-10-29
Unify permissions

leonie-koch:
This commit streamlines permission handling in AuthService by making it less restrictive, especially for pending documentation units. Both the creating and owning documentation offices can now edit and delete documents. While the UI ensures the owning office cannot enter edit mode before a takeover, this change reduces code complexity (e.g. the status and docUnitIsPending checks are not needed anymore.)

Additionally, hasWriteAccess permissions have been unified. Beans like userHasSameDocOfficeAsDocument were removed since they are now covered by hasWriteAccess.

This should improve code readability and may also enhance performance by reducing redundant checks.


HPrinz:
Looks good! Do we know to what extent this is covered by tests? Maybe we can setup a few unit tests for the OAuthService?

PR #2323, opened 2024-10-31
Improve visibility of local env by using rainbow

anne-ds:
As the favicon looks exactly the same in local and staging environment it happens that you mix up the browser tabs. With this change this will not happen anymore.

Left after, right before...
image


kiwikern (approved):
Ich liebs!!! 😍 🌈

PR #2328, opened 2024-11-01
Refactor container el ref for ListInput

kiwikern:
RISDEV-5178
Vue offers an API, so we don't need to do it manually.


anne-ds (approved):
🆒 !

PR #2334, opened 2024-11-04
Delete reference with Neuanlage

HPrinz:


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/PostgresLegalPeriodicalEditionRepositoryImpl.java
```diff
@@ -51,26 +51,33 @@ public LegalPeriodicalEdition save(LegalPeriodicalEdition legalPeriodicalEdition
...
-        }
-        // delete all deleted references
-        documentationUnitRepository
-            .findById(reference.getDocumentationUnit().getId())
-            .ifPresent(
-                docUnit -> {
-                  docUnit.getReferences().remove(reference);
-                  documentationUnitRepository.save(docUnit);
-                });
+  private void deleteDocUnitLinksForDeletedReferences(LegalPeriodicalEdition updatedEdition) {
```
Nadav-B, line 54: Alternatively, we could add a `deleteReference` endpoint in the controller specifically for `Reference` that only takes an `id`. This way, we could have avoided the filtering out and start directly by the desired item, not having to send the entire list of references on deletion.

PR #2341, opened 2024-11-06
Add edit and remove table in editor

anne-ds:
[RISDEV-4121](https://digitalservicebund.atlassian.net/browse/RISDEV-4121)

[RISDEV-4121]: https://digitalservicebund.atlassian.net/browse/RISDEV-4121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/src/components/input/TextEditorMenu.vue
```diff
@@ -192,6 +198,65 @@ const buttons = computed(() => {
...
+      group: "Tabelle",
+      isCollapsable: true,
+      isAlwaysCollapsed: true,
+      callback: () =>
+        props.editor
+          .chain()
+          .focus()
+          .insertTable({ rows: 3, cols: 3, withHeaderRow: false })
+          .run(),
+    },
```
kiwikern, line 214: blocking: I understand that we do not want to test the library itself, however, we should test that our integration with the library works. Let's imagine the lib's new major version introduces a `commit()` method that needs to be called after `run()`. Without calling it, nothing happens. Renovate would just update the lib and the pipeline would be green.

I would argue for a minimal e2e test case like this:
* Uncollapse table menu -> check all buttons are displayed
* Click add table button and check that a table element is in editor (no details about table)

frontend/src/editor/tableStyle.ts
```diff
@@ -1,10 +1,41 @@
...
+                  !trimmedStyle.includes("none")
+                )
+              })
+
+              const invisibleTableClass = hasBorderStyle
+                ? ""
+                : "invisible-table"
+
+              return {
+                class: invisibleTableClass,
```
kiwikern, line 30: question: Could there be an existing `class` that we would overwrite?

frontend/src/styles/tables.css
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  .tableWrapper {
+    margin: 1.5rem 0;
+    overflow-x: auto;
+  }
+
+  &.resize-cursor {
+    cursor: ew-resize;
+    cursor: col-resize;
```
kiwikern, line 72: question: Why do we have two cursors here? Should `ew-resize` be removed?

frontend/src/styles/tables.css
kiwikern, line 1: out-of-scope:
These styles are quite extensive, it's impressive how many cases you have covered. Here, I feel like we are missing visual regression testing a little bit. Conditionally adding the invisible table style is untested. 🤔 Something to keep in mind, in case we develop a component lib between the streams and add visual regression testing then.

kiwikern (commented):
Really cool how quickly you added this feature. Although the library gives us a lot, there were still a lot of smaller things to attend to. Great job! 💪

frontend/src/components/input/TextEditorButton.vue
```diff
@@ -56,6 +56,7 @@ export interface EditorButton {
   isLast?: boolean
   isActive?: boolean
   isCollapsable?: boolean
+  isAlwaysCollapsed?: boolean
```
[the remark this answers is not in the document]
kiwikern: Oh, just saw that you removed it in a later commit. So when you have children but `isCollapsable: false` it is always collapsed? This is a bit confusing, but apparently it was like this before.

PR #2353, opened 2024-11-11
Strengthening recycling service conditions

Nadav-B:
RISDEV-5186

* Removing optional from service to throw actual errors
* Adding and adjusting tests
* addForRecycling to void
* Adding custom exception
* Empty log in either handover or migration service is required to for recycle. This condition is a must on both recycling and saving

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
@@ -173,7 +173,6 @@ public class DatabaseDocumentNumberRecyclingService implements DocumentNumberRec
    * @throws DocumentNumberRecyclingException will through error if published / handed over or
    *     status history not empty.
    */
-  @Override
   public void assertStatusHasNeverBeenPublished(DocumentationUnitDTO documentationUnitDTO)
       throws DocumentNumberRecyclingException {
     PublicationStatus status = documentationUnitDTO.getStatus().getPublicationStatus();
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
@@ -80,7 +80,7 @@ public class DatabaseDocumentNumberRecyclingService implements DocumentNumberRec
       repository.save(deleted);
 
     } catch (Exception e) {
-      log.warn("Won´t reuse the document number", e);
+      log.info("Won´t reuse the document number", e);
       throw e;
     }
   }
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
@@ -5,68 +5,70 @@ import static de.bund.digitalservice.ris.caselaw.domain.DateUtil.getYear;
 import de.bund.digitalservice.ris.caselaw.adapter.database.jpa.DatabaseDeletedDocumentationIdsRepository;
 import de.bund.digitalservice.ris.caselaw.adapter.database.jpa.DatabaseDocumentationUnitRepository;
 import de.bund.digitalservice.ris.caselaw.adapter.database.jpa.DeletedDocumentationUnitDTO;
+import de.bund.digitalservice.ris.caselaw.adapter.database.jpa.DocumentationUnitDTO;
 import de.bund.digitalservice.ris.caselaw.domain.DocumentNumberRecyclingService;
+import de.bund.digitalservice.ris.caselaw.domain.HandoverEntityType;
+import de.bund.digitalservice.ris.caselaw.domain.HandoverService;
 import de.bund.digitalservice.ris.caselaw.domain.PublicationStatus;
 import de.bund.digitalservice.ris.caselaw.domain.exception.DocumentNumberPatternException;
+import de.bund.digitalservice.ris.caselaw.domain.exception.DocumentNumberRecyclingException;
 import de.bund.digitalservice.ris.caselaw.domain.exception.DocumentationUnitNotExistsException;
 import java.time.Year;
-import java.util.Optional;
 import java.util.UUID;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 
-/** Service to keep track off deleted unpublished documentation units that can be recycled */
+/** Service to handle deleted unpublished documentation units that can be recycled */
 @Service
 @Slf4j
 public class DatabaseDocumentNumberRecyclingService implements DocumentNumberRecyclingService {
 
   private final DatabaseDeletedDocumentationIdsRepository repository;
+  private final DatabaseDocumentationUnitRepository documentationUnitRepository;
+
   private final DocumentNumberPatternConfig documentNumberPatternConfig;
 
-  private final DatabaseDocumentationUnitRepository documentationUnitRepository;
+  private final HandoverService handoverService;
 
   public DatabaseDocumentNumberRecyclingService(
       DatabaseDeletedDocumentationIdsRepository repository,
       DocumentNumberPatternConfig documentNumberPatternConfig,
+      HandoverService handoverService,
       DatabaseDocumentationUnitRepository documentationUnitRepository) {
     this.repository = repository;
     this.documentNumberPatternConfig = documentNumberPatternConfig;
+    this.handoverService = handoverService;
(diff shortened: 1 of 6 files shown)


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
```diff
@@ -75,45 +77,110 @@ public Optional<String> addForRecycling(
...
       var deletedDocumentationUnitDTO = optionalDeletedDocumentationUnitDTO.get();
 
-      if (!documentNumberPatternConfig.hasValidPattern(
-          documentationOfficeAbbreviation, deletedDocumentationUnitDTO.getDocumentNumber())) {
-        throw new DocumentNumberPatternException("Pattern is invalid");
+      if (documentationUnitRepository
+          .findByDocumentNumber(deletedDocumentationUnitDTO.getDocumentNumber())
+          .isPresent()) {
+        throw new DocumentNumberRecyclingException(
+            "Saved documentation number is currently in use");
```
HPrinz, line 116: Hey @Nadav-B - This is has not directly been your task, but I just saw that in this case, we could(should) probably remove the entry from DatabaseDeletedDocumentationIdsRepository

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
```diff
@@ -75,45 +77,110 @@ public Optional<String> addForRecycling(
...
+   *     status history not empty.
+   */
+  public void assertStatusHasNeverBeenPublished(DocumentationUnitDTO documentationUnitDTO)
+      throws DocumentNumberRecyclingException {
+    PublicationStatus status = documentationUnitDTO.getStatus().getPublicationStatus();
+    if (documentationUnitDTO.getStatusHistory().size() != 1
+        || !(status.equals(PublicationStatus.UNPUBLISHED)
+            || status.equals(PublicationStatus.EXTERNAL_HANDOVER_PENDING))) {
+      throw new DocumentNumberRecyclingException(
+          "Status has been changed once or neither unpublished nor handover pending");
```
HPrinz, line 183: Hey @Nadav-B - Do you think we would safely remove this check? From my understanding, status history does not tell us anything about potential sync problems with jDV because of document number recycling. It's not harmful either, but to prevent misunderstandings and keep the amount of code to a minimum, I lean towards deletion

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
```diff
@@ -75,45 +77,110 @@ public Optional<String> addForRecycling(
...
+          .isPresent()) {
+        throw new DocumentNumberRecyclingException(
+            "Saved documentation number is currently in use");
       }
-      return Optional.of(deletedDocumentationUnitDTO.getDocumentNumber());
+
+      assertPatternIsValid(
+          documentationOfficeAbbreviation, deletedDocumentationUnitDTO.getDocumentNumber());
+
+      return deletedDocumentationUnitDTO.getDocumentNumber();
```
HPrinz, line 122: This method is much easier to read now 👍

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
```diff
@@ -5,68 +5,70 @@
...
 
-      PublicationStatus status = docUnit.getStatus().getPublicationStatus();
-      if (docUnit.getStatusHistory().size() != 1
-          || !(status.equals(PublicationStatus.UNPUBLISHED)
-              || status.equals(PublicationStatus.EXTERNAL_HANDOVER_PENDING))) {
-        throw new DocumentNumberPatternException(
-            "Status is changed or neither unpublished nor pending");
-      }
+      assertDocumentationUnitHasNeverBeenHandedOverOrMigrated(documentationUnitId);
+
```
HPrinz, line 70: Yay, much better than before and also properly documented 🏅

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
```diff
@@ -75,45 +77,110 @@ public Optional<String> addForRecycling(
               .abbreviation(documentationOfficeAbbreviation)
               .build();
 
-      return Optional.of(repository.save(deleted).getDocumentNumber());
+      repository.save(deleted);
 
     } catch (Exception e) {
       log.warn("Won´t reuse the document number", e);
```
HPrinz, line 83: @Nadav-B : This has not been your task, but in case you're editing the file, we could change this to log.info. don't you think?

Nadav-B: I hope I understand you correctly, but `addForRecycling` is only called on deletion of documentation unit.

HPrinz: Yes, exactly! And when a docunit is deleted and the number can't be recycled, it would count as expected behaviour, right?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
HPrinz: I can't make up a scenario where the check for an empty event log would not be sufficient 🤔. And yes, we have been saving the original XML of each decision since day 1 and this is included into the event log. But maybe it's worth getting a third opinion on this?

Limiting the recycling of document numbers for the current year is a good idea for optimization, but since you check when the actual recycling happens, I don't think it's worth the effort.

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
Nadav-B: Oh I see you mean in line 81, I was confused because of the marking. [Done](https://github.com/digitalservicebund/ris-backend-service/commit/208b683f12d7cb7c20bc50431bc3017e0096b90c#diff-ac6fb6e22dffe44d234f1e5390569a0dff92c4a584f7283dba599c9fcd4a55a4)

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
Nadav-B: @rvp-c Long story short: we used the status history to determine rather we can recycle a document number. Hanna had a good point to rely only on the event log (migration and handover), meaning as long as it was never published or migrated to jdv we can reuse it. 

This check is now added on top. However, another point is to remove the check in the `getLatestStauts` and `statusHistory` for unpublished, as it might not be needed. Do you see any problem with it or shall we adapt @HPrinz idea and rely on the event log only?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
Nadav-B: @HPrinz In line [125](https://github.com/digitalservicebund/ris-backend-service/blob/65a7432edfa044f3368323dfa727a1db08bd087b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java#L126-L126), it will end up removing the deleted document in the catch exception.
```
 optionalDeletedDocumentationUnitDTO.ifPresent(
          deletedDocumentationUnitDTO ->
              repository.deleteById(deletedDocumentationUnitDTO.getDocumentNumber()));
```

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
Nadav-B: @HPrinz Ralph's feedback has no objection to your idea. However as it is not effecting the code, I will put the ticket, and if it is required we extract this to a new ticket.

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDocumentNumberRecyclingService.java
HPrinz: Oh, haven't seen that, thank you! 🙏

PR #2362, opened 2024-11-14
Add missing inputs for literature references

leonie-koch:
- new document type controller endpoint to get all document types with category "U"
- added radio buttons for conditional rendering of literature or caselaw reference inputs
- both for periodical evaluation reference input and documentation unit reference input

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/PostgresDocumentTypeRepositoryImpl.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/PostgresDocumentTypeRepositoryImpl.java
@@ -31,6 +31,25 @@ public class PostgresDocumentTypeRepositoryImpl implements DocumentTypeRepositor
   }
 
   @Override
+  public List<DocumentType> findIndependentLiteratureBySearchStr(String searchString) {
+    return repository
+        .findCaselawBySearchStrAndCategory(
+            searchString, categoryRepository.findFirstByLabel("U").getId())
+        .stream()
+        .map(DocumentTypeTransformer::transformToDomain)
+        .toList();
+  }
+
+  @Override
+  public List<DocumentType> findAllIndependentLiteratureOrderByAbbreviationAscLabelAsc() {
+    return repository
+        .findAllByCategoryOrderByAbbreviationAscLabelAsc(categoryRepository.findFirstByLabel("U"))
+        .stream()
+        .map(DocumentTypeTransformer::transformToDomain)
+        .toList();
+  }
+
+  @Override
   public Optional<DocumentType> findUniqueCaselawBySearchStr(String searchString) {
     return repository
         .findUniqueCaselawBySearchStrAndCategory(
--- a/frontend/src/components/DocumentUnitReferenceInput.vue
+++ b/frontend/src/components/DocumentUnitReferenceInput.vue
@@ -1,7 +1,7 @@
 <script lang="ts" setup>
 import { computed, ref, watch } from "vue"
 import ComboboxInput from "@/components/ComboboxInput.vue"
-import InputField from "@/components/input/InputField.vue"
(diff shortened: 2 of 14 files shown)


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/PostgresDocumentTypeRepositoryImpl.java
```diff
@@ -30,6 +30,25 @@ public List<DocumentType> findCaselawBySearchStr(String searchString) {
...
+        .toList();
+  }
+
+  @Override
+  public List<DocumentType> findAllIndependentLiteratureOrderByAbbreviationAscLabelAsc() {
+    return repository
+        .findAllByCategoryOrderByAbbreviationAscLabelAsc(categoryRepository.findFirstByLabel("U"))
+        .stream()
+        .map(DocumentTypeTransformer::transformToDomain)
+        .toList();
```
HPrinz, line 49: @leonie-koch Just one note on the method names: I believe U=Unselbstständige Literatur would be DependentLiterature (~Abhängige Literatur) instead of Independent, right?

frontend/src/components/DocumentUnitReferenceInput.vue
```diff
@@ -86,6 +87,41 @@ watch(
 
 <template>
   <div class="flex flex-col gap-24">
+    <div v-if="lastSavedModelValue.isEmpty" class="flex items-center gap-16">
+      <div class="flex items-center">
+        <InputField
+          id="caselaw"
```
HPrinz, line 93: @leonie-koch Should we give it a more precise name so that we make sure the id is unique? Like `caselaw-reference-type`

frontend/src/components/periodical-evaluation/references/PeriodicalEditionReferenceInput.vue
```diff
@@ -297,261 +298,331 @@ onMounted(async () => {
...
+              ></TextInput>
+              <TextInput
+                id="citation"
+                v-model="reference.citation"
+                aria-label="Zitatstelle *"
+                :has-error="slotProps.hasError"
+                placeholder="Variable"
+                size="medium"
+                @blur="validateRequiredInput(reference)"
+                @focus="validationStore.remove('citation')"
```
HPrinz, line 373: @leonie-koch When testing, I noticed that the required input for literature references (including citation) are not validated correctly. However, in the code it looks all good

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/PostgresDocumentTypeRepositoryImpl.java
leonie-koch: absolutely right, this translation really messed with my mind (INdependent != UNabhängig) :D i fixed that

frontend/src/domain/reference.ts
Nadav-B, line 1: Hey @leonie-koch, firstly nice job for this big commit implementation. 

I have a small suggestion regarding the Reference. I would have to extract Reference class to an interface and then extend to `LiteratureReference` and `CaselawReference`.  This way we can assign different required fields with a clearer code separation to both classes. Keeping exactly one requiredFields to each class, like we have been doing with our other types. 

So for instance [requiredFields](https://github.com/digitalservicebund/ris-backend-service/blob/6cf0bf3fa3a430104a526fb41682ed336f19db83/frontend/src/domain/reference.ts#L18) can exactly be assign to our types. 

Alternatively, if we do keep with Reference as one, the [validationStore](https://github.com/digitalservicebund/ris-backend-service/blob/6cf0bf3fa3a430104a526fb41682ed336f19db83/frontend/src/domain/reference.ts#L65) in Reference needed to be adjusted accordingly as it now validates on `addReaddReference` referenceSuppliment for lietarture fundstelle. 
We might need a new requiredFields specific for literature.

PR #2370, opened 2024-11-18
RISDEV-5240

HPrinz:


frontend/src/domain/reference.ts
```diff
@@ -50,9 +50,11 @@ export default class Reference implements EditableListItem {
 
   get renderDecision(): string {
     return [
+      this.author ? `${this.author},` : "",
       this.legalPeriodical?.abbreviation ?? this.legalPeriodicalRawValue,
       this.citation,
       this.referenceSupplement ? ` (${this.referenceSupplement})` : "",
+      this.documentType ? ` (${this.documentType.jurisShortcut})` : "",
```
HPrinz, line 57: @leonie-koch For some reason, this does not work for saved decisions. Looks a bit like the objects returned by the servie are not of the right type (at least I had similar problems with that)

HPrinz: Same with the Author

frontend/src/components/periodical-evaluation/references/PeriodicalEditionReferenceInput.vue
```diff
@@ -174,7 +179,11 @@ function addReference(decision: RelatedDocumentation) {
 
   validateRequiredInput(newReference)
 
-  if (validationStore.isValid()) {
+  // Todo validate also literature input
+  if (
+    validationStore.isValid() ||
+    reference.value.referenceType == "literature"
```
HPrinz, line 185: @leonie-koch Ah, I found the reason for [this](https://github.com/digitalservicebund/ris-backend-service/pull/2370/files#r1846555995)

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/transformer/DependentLiteratureTransformer.java
```diff
@@ -0,0 +1,73 @@
...
+        .id(reference.id())
+        .author(reference.author())
+        .citation(reference.citation())
+        .documentType(documentType)
+        .legalPeriodicalRawValue(
+            legalPeriodicalRawValue != null
+                ? legalPeriodicalRawValue
+                : reference.legalPeriodicalRawValue())
+        .legalPeriodical(legalPeriodicalDTO)
+        .type(DependentLiteratureCitationType.PASSIVE)
```
HPrinz, line 68: @leonie-koch I think this should be an active citation, right?

frontend/src/domain/reference.ts
HPrinz: Ah, I think I finally understood the problem. In the backend, we need to save literatureReferences separately from the references. Right now, they are saved as normal references and therefore the author and documentType are always `null`

frontend/test/e2e/caselaw/legal-periodical-evaluation.spec.ts
```diff
@@ -576,17 +577,37 @@ test.describe(
...
-        // })
+          await searchForDocUnitWithFileNumber(page, fileNumber, "31.12.2019")
+          await page.getByLabel("Treffer übernehmen").click()
+          await expect(
+            page.getByText("Bilen, Ulviye, MMG 2024, 2021, 2, Heft 1 (Ean)"),
+          ).toBeVisible()
+        })
 
-        // await test.step("Literature reference correctly sorted with other references", async () => {
+        // await test.step("Literature reference are shown in correct order", async () => {
```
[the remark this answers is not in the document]
HPrinz: Nevermind. It works but for the wrong reasons (literature references are not saved in the dedicated table)

PR #2390, opened 2024-11-22
Improve search query performance

kiwikern:
RISDEV-5446
For the query optimizer to be able to use the index, we cannot use dynamic conditions in the CASE WHEN statements in the order by statement.

With this change, the index is used for queries that do not relate to Terminierte Abgabe.


anne-ds (approved):
Thanks a lot for fixing my weird code. I keep my fingers crossed this helps speeding up the search again.🤞 🍀 🍀 🍀 🤞

PR #2391, opened 2024-11-22
Improve search query performance

kiwikern:
RISDEV-0000
These additional categories are not relevant for the search results but are still fetched for each row (N+1 problem) because of the EAGER fetching. With lazy fetching, they will only be queried on demand (e.g. when loading the whole doc unit).


kiwikern:
For each row that is fetching during a search, we do 9 SQL queries. So for fetching a search that has 100 results, we run 900 SQL queries against the database. With this change, we reduce the number of queries by 5 per row.

anne-ds (approved):

PR #2407, opened 2024-11-29
Notify about handover errors

kiwikern:


backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/ScheduledPublicationService.java
```diff
@@ -36,7 +36,8 @@ public ScheduledPublicationService(
     this.mailSender = mailSender;
   }
 
-  @Scheduled(fixedRateString = "PT1M") // Runs every minute
+  // Runs every minute, starts initially after 5s
+  @Scheduled(fixedRateString = "PT1M", initialDelayString = "PT5S")
```
kiwikern, line 40: We now start the job right after the server starts (with a 5s delay). To me this makes sense in general and also reduces the integration test duration from 60s to 5s.

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/ScheduledPublicationService.java
```diff
@@ -53,7 +54,10 @@ public void handoverScheduledDocUnits() {
   private void handoverDocument(DocumentationUnit docUnit) {
     try {
       String email = docUnit.managementData().scheduledByEmail();
-      this.handoverService.handoverDocumentationUnitAsMail(docUnit.uuid(), email);
+      var result = this.handoverService.handoverDocumentationUnitAsMail(docUnit.uuid(), email);
+      if (!result.isSuccess()) {
+        throw new HandoverException(String.join(", ", result.statusMessages()));
+      }
```
kiwikern, line 60: This is also what the /handover controller endpoint does. This way, we do not have to change the logic of the handover method, but still send out the error notification.

PR #2426, opened 2024-12-11
Cancel+debounce Combobox requests + delayed enter

kiwikern:
RISDEV-5663
Previously: Requests on every keystroke. Now, we debounce and cancel requests. Also, when enter is pressed while a search is ongoing, we defer the enter to select the first item to when the request resolves.

### DIFF
--- a/frontend/src/components/ComboboxInput.vue
+++ b/frontend/src/components/ComboboxInput.vue
@@ -151,15 +151,21 @@ const updateFocusedItem = () => {
 
 const isFetchingOrTyping = ref(false)
 
+/**
+ * When a new request is started while a previous one is still running -> cancel the old one.
+ * The canceled one should not unset the isFetching state as it is followed by a new request.
+ */
 async function updateCurrentItems() {
   isFetchingOrTyping.value = true
   if (canAbort.value) {
     abort()
   }
   const response = await debouncedFetchItems()
-  if (response) isFetchingOrTyping.value = false
+  const wasCanceled = !response
+  if (!wasCanceled) isFetchingOrTyping.value = false
 }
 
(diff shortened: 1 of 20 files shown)


frontend/src/components/ComboboxInput.vue
```diff
@@ -149,65 +149,63 @@ const updateFocusedItem = () => {
...
-    Sentry.captureMessage(
-      "more than 5 tries to call the item service at the combobox",
-      "error",
-    )
+async function updateCurrentItems() {
+  isFetchingOrTyping.value = true
+  if (canAbort.value) {
+    abort()
   }
+  const response = await debouncedFetchItems()
```
anne-ds, line 159: const wasCanceled = !response

frontend/test/components/comboboxInput.spec.ts
```diff
@@ -15,61 +45,45 @@ function renderComponent(
...
+    expect(screen.queryByDisplayValue("courtlabel1")).not.toBeInTheDocument()
+    expect(screen.queryByDisplayValue("courtlabel1")).not.toBeInTheDocument()
   })
 
   it("is opened", async () => {
     renderComponent()
 
     const openComboboxContainer = screen.getByLabelText("Dropdown öffnen")
     await user.click(openComboboxContainer)
+    await vi.advanceTimersByTimeAsync(300)
```
anne-ds, line 86: 200

frontend/test/components/comboboxInput.spec.ts
```diff
@@ -78,28 +92,34 @@ describe("Combobox Element", () => {
   })
 
   it("focus should open dropdown", async () => {
+    vi.useFakeTimers()
```
anne-ds, line 95: delete

PR #2433, opened 2024-12-17
Limit combobox court results to 200

kiwikern:
RISDEV-5901
Doc offices complained that 10 is not enough to find matching courts.


HPrinz (approved):
Looks all good - thanks @kiwikern!

PR #2446, opened 2024-12-23
Use directive to trigger search with ctrl + enter shortcut in all searches

HPrinz:
I added a [vue directive](https://vuejs.org/guide/reusability/custom-directives.html) that allows components to react to `ctrl` + `enter` shortcuts. The action is only triggered if it the shortcut is pressed within the active/focused component. This makes sure that, when the user enters data in active citation, only the active citation small search is triggered but not the previous decision search (see the if clause in handleKeydown()) --> [link to file](https://github.com/digitalservicebund/ris-backend-service/blob/bddfd7b0129ec6e4a51fee77aabea8d02454decd/frontend/src/utils/searchShortcutDirective.ts)

```
import { Directive } from "vue"

export const searchShortcutDirective: Directive = {
  mounted(el, binding) {
    const handleKeydown = (event: KeyboardEvent) => {
      if (
        el.contains(document.activeElement) && // Ensure focus is inside the element
        event.key === "Enter" &&
        (event.ctrlKey || event.metaKey)
      ) {
        binding.value?.(event) // Call the provided callback
      }
    }

    el.__handleKeydown__ = handleKeydown // Store reference for cleanup
    window.addEventListener("keydown", handleKeydown)
  },
  unmounted(el) {
    window.removeEventListener("keydown", el.__handleKeydown__)
    delete el.__handleKeydown__
  },
}
```

I also replaced the current implementation of the shortcut search in the big search with the directive solution --> see [here](https://github.com/digitalservicebund/ris-backend-service/pull/2446/files#diff-4d0b0fe2a21fe00ac9c333ac2e498c9037d26d8401cfd9de27ff8b8d97136144)

Most of the code changes are due to adding the directive into the frontend tests and adding tests themselves :)

### DIFF
--- a/frontend/src/components/ActiveCitationInput.vue
+++ b/frontend/src/components/ActiveCitationInput.vue
@@ -212,7 +212,7 @@ onMounted(() => {
 </script>
 
 <template>
-  <div class="flex flex-col gap-24">
+  <div v-search="search" class="flex flex-col gap-24">
     <InputField
       id="activeCitationPredicate"
       v-slot="slotProps"
--- a/frontend/src/components/ActiveCitationInput.vue
+++ b/frontend/src/components/ActiveCitationInput.vue
@@ -212,7 +212,7 @@ onMounted(() => {
 </script>
 
 <template>
-  <div v-search="search" class="flex flex-col gap-24">
+  <div v-ctrl-enter="search" class="flex flex-col gap-24">
     <InputField
       id="activeCitationPredicate"
       v-slot="slotProps"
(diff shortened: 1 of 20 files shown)


leonie-koch (commented):
Wow!! Das ist wirklich eine tolle Idee, wusste nicht dass das geht. das einzige was mir dabei auffällt: bei v-search denkt man sich sofort daran, dass man hier den callback für einen shortcut hinterlegen kann. Wollen wir da im naming irgendwie klarer machen, dass es sich hierbei um den ctrl + enter callback handelt?

PR #2458, opened 2025-01-02
RISDEV-5906

leonie-koch:
Automatically open preview, when only one search result was found.
This seems to add flaky behaviour to the tests, that's why the long test file was reorganized in smaller steps to better understand the issue.


leonie-koch:
already in main

PR #2461, opened 2025-01-02
RISDEV-5904

leonie-koch:
Clicking on a document number opens up the preview, scrolled to "Leitsatz", if present. If not, it scrolls to the top of the document.


leonie-koch:
already on main

PR #2464, opened 2025-01-06
Reset document number count on each year

Nadav-B:
RISDEV-6004

* Adding year to document number table
* Dropping old constraint on abbreviation and update on id
* Adding combined index for year and doc office
* Backfilling 2024 for former year entries
* Retrieve document number by both doc office and year
* Extracting getOrCreateDocumentNumberDTO to method
* Adding tests

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/DatabaseDocumentNumberRepository.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/DatabaseDocumentNumberRepository.java
@@ -1,15 +1,13 @@
 package de.bund.digitalservice.ris.caselaw.adapter.database.jpa;
 
-import jakarta.persistence.LockModeType;
+import java.time.Year;
 import java.util.Optional;
-import org.jetbrains.annotations.NotNull;
 import org.springframework.data.jpa.repository.JpaRepository;
-import org.springframework.data.jpa.repository.Lock;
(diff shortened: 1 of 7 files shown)


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/DatabaseDocumentNumberRepository.java
```diff
@@ -1,15 +1,13 @@
...
 import java.util.Optional;
-import org.jetbrains.annotations.NotNull;
 import org.springframework.data.jpa.repository.JpaRepository;
-import org.springframework.data.jpa.repository.Lock;
 import org.springframework.stereotype.Repository;
 
 @Repository
 public interface DatabaseDocumentNumberRepository extends JpaRepository<DocumentNumberDTO, String> {
-  @NotNull
-  @Lock(LockModeType.PESSIMISTIC_WRITE)
```
rvp-c, line 13: I think this is something we need also at the new function

rvp-c: The idea behind this is to allow only one thread to get a document number at a time. All other threads have to wait until the first thread is ready. If we don't do this two thread can get the same document number and the unique constrain will reject the second one to generate a new documentation unit.

rvp-c (changes requested):

rvp-c (approved):

PR #2494, opened 2025-01-15
Lefthook manual command

HPrinz:


lefthook.yaml
```diff
@@ -87,6 +87,26 @@ pre-push:
           exit 1
         fi
 
+tests:
+  follow: true
```
HPrinz, line 91: @Nadav-B I added this config to log the test output during execution. Of course the lines of all three processes will be mixed, but still some real-time feedback. Let me know what you think!

lefthook.yaml
```diff
@@ -87,6 +87,26 @@ pre-push:
...
+  parallel: true
+  commands:
+    frontend-tests:
+      tags: frontend tests
+      root: frontend/
+      run: npm run check --verbose 2>&1
```
HPrinz, line 97: Not sure we need the --verbose option?

lefthook.yaml
```diff
@@ -87,6 +87,19 @@ pre-push:
...
+  parallel: true
+  commands:
+    frontend-tests:
+      tags: frontend tests
+      root: frontend/
+      run: npm run check 2>&1
+    backend-unit-tests:
+      tags: backend tests
+      root: backend/
+      run: ./gradlew test -x integrationTest 2>&1
```
HPrinz, line 101: I removed the plain output option here - ok @Nadav-B ?

lefthook.yaml
```diff
@@ -87,6 +87,19 @@ pre-push:
...
+    backend-integration-tests:
+      tags: backend tests
+      root: backend/
+      run: ./gradlew integrationTest -x test 2>&1
```
HPrinz, line 105: Here as well

PR #2503, opened 2025-01-20
Add duplicate check e2e

kiwikern:
RISDEV-5883

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DocumentationUnitController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DocumentationUnitController.java
@@ -496,6 +496,11 @@ public class DocumentationUnitController {
       @PathVariable String documentNumberDuplicate,
       @RequestBody @Valid DuplicateRelationStatusRequest duplicateRelationStatusRequest) {
     try {
+      log.info(
+          "PUT duplicate relation status: {}, {}, {}",
+          documentNumberOrigin,
+          documentNumberDuplicate,
+          duplicateRelationStatusRequest.getStatus());
       var result =
           duplicateCheckService.updateDuplicateStatus(
               documentNumberOrigin,
@@ -503,6 +508,7 @@ public class DocumentationUnitController {
               duplicateRelationStatusRequest.getStatus());
       return ResponseEntity.status(HttpStatus.OK).body(result);
     } catch (DocumentationUnitNotExistsException | EntityNotFoundException ex) {
+      log.error("Could not find given relationship", ex);
       return ResponseEntity.notFound().build();
     }
   }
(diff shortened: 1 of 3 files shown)


frontend/test/e2e/caselaw/management-data/duplicate-check.spec.ts
```diff
@@ -0,0 +1,415 @@
...
+        await navigateToManagementData(page, prefilledDocNumber)
+        await expectNoDuplicateWarning(page)
+        await test.step("Reset duplicate warning", async () => {
+          await page.getByLabel("Warnung ignorieren").uncheck()
+        })
+        await expectDuplicateWarning(page)
+      })
+    })
+
+    // FIXME: ignoring second duplicate warning is flaky
```
kiwikern, line 135: TODO: Test is flaky

frontend/test/e2e/caselaw/management-data/duplicate-check.spec.ts
anne-ds, line 1: You called the folder "management-data". I was just wondering what our naming pattern was when I realized that we actually have both camel case (more often) and kebap-case... 🙈

frontend/test/e2e/caselaw/management-data/duplicate-check.spec.ts
```diff
@@ -0,0 +1,415 @@
...
+        await expectDuplicateWarning(page)
+      })
+
+      await test.step("Check management data and ignore warning", async () => {
+        await navigateToManagementData(page, documentNumber)
+        await expectDocUnitSummaryInDuplicatesList(page, prefilledDocumentUnit)
+        await ignoreDuplicateWarning(page, prefilledDocumentUnit)
+        await expectNoDuplicateWarning(page)
+
+        await test.step("Duplicate warning should be ignored on other doc unit as well", async () => {
```
anne-ds, line 119: I don't understand the description of this step. What you actually check is if the checkbox of the same doc unit is still checked? 🤔

anne-ds (approved):

frontend/test/e2e/caselaw/management-data/duplicate-check.spec.ts
kiwikern: As I think it is better to avoid casing in file/folder names and our route is `managementdata` (no minus) anyway, I just renamed it to `managementdata`. 😅

frontend/test/e2e/caselaw/management-data/duplicate-check.spec.ts
kiwikern: That was wrong, thanks for catching! 🦅 👁️ 
I moved it downward where the other side is tested.

frontend/test/e2e/caselaw/management-data/duplicate-check.spec.ts
kiwikern: Locally it's of course not flaky 😢 but when running on bulk against staging, it seems to be 20% failure rate. :-/

PR #2519, opened 2025-01-23
Add delete button to managementdata

kiwikern:
RISDEV-5885


frontend/src/components/DocumentUnitDeleteButton.vue
```diff
@@ -0,0 +1,54 @@
...
+    `Möchten Sie die Dokumentationseinheit ${documentNumber} wirklich dauerhaft löschen?`,
+)
+const router = useRouter()
+
+const deleteDocumentUnit = async () => {
+  if (!uuid) return
+
+  const { error, data } = await documentUnitService.delete(uuid)
+  if (error) {
+    alert("Fehler beim Löschen der Dokumentationseinheit: " + data)
```
anne-ds, line 30: In the design there is a modal not an alert. We might use this modal also in the search results list (consistency).

frontend/src/components/DocumentUnitDeleteButton.vue
anne-ds, line 1: Would it be possible to have only one component for the Deletion that we could also reuse in the search result list?

frontend/src/components/DocumentUnitDeleteButton.vue
anne-ds, line 30: Should we close the other modal in that case?

frontend/src/components/DocumentUnitDeleteButton.vue
kiwikern: I just began, but it needs a bit of conceptual work: 
1. The lists do not need navigation after the deletion, but instead want to remove the items from the list results without reloading search results or procedures.
2. The lists currently have "one dialog element" that is triggered by the button with different params. If we followed the approach, we would have one dialog element for every list item. This feels strange to me. I think we should put more conceptual work in our dialog component, should rather work like a singleton. (It does not make sense to show multiple dialogs at the same time, same with snackbar or similar.)

--> I would postpone this, although it would be very nice to do this, especially as the logic in the lists does not seem great and the error dialog is much better than the alert.

PR #2529, opened 2025-01-27
Duplicate check cron job

kiwikern:
RISDEV-5912


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDuplicateCheckService.java
```diff
@@ -67,6 +69,16 @@ public void checkDuplicates(String docNumber) {
     }
   }
 
+  // Runs every night at 05:05:10
+  //  @Scheduled(cron = "10 5 5 * * *")
+  @Scheduled(fixedRateString = "PT1M")
```
anne-ds, line 74: The comment says every night, but the value represents every month?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/DuplicateRelationViewRepository.java
```diff
@@ -0,0 +1,64 @@
...
+       drel.id_b,
+       CAST(CASE
+                WHEN d1.duplicate_check = FALSE OR d2.duplicate_check = FALSE THEN 'IGNORED'
+                ELSE 'PENDING' END AS incremental_migration.duplicate_relation_status)
+FROM incremental_migration.duplicate_relations_view drel
+         LEFT JOIN incremental_migration.duplicate_relation ON drel.id_a = duplicate_relation.documentation_unit_id1 AND
+                                         drel.id_b = duplicate_relation.documentation_unit_id2
+         LEFT JOIN incremental_migration.documentation_unit d1 ON drel.id_a = d1.id
+         LEFT JOIN incremental_migration.documentation_unit d2 ON drel.id_b = d2.id
+WHERE duplicate_relation.documentation_unit_id1 IS NULL;
```
anne-ds, line 46: Why is this condition sufficient and 'AND duplicate_relation.**documentation_unit_id2** IS NULL' is not needed?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/DatabaseDuplicateCheckService.java
kiwikern: Sorry, das ist nur fürs Testen, das läuft minütlich. Das auskommentierte darüber wäre einmal nachts um 05:05:10

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/DuplicateRelationViewRepository.java
anne-ds, line 1: There is a lot of logic in here. As far as I can see, this has not yet been tested?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/DuplicateRelationViewRepository.java
kiwikern: Checking one side is enough, because for a duplicate relationship, both ids will always be set if present. As we are doing a `LEFT JOIN`, we get all rows from the materialized view with existing and non-existing duplicate relationships. By filtering for NULL, we only get non-existing relationships.
image

PR #2533, opened 2025-01-28
Attempt to fix norms tests

HPrinz:
RISDEV-5965


leonie-koch:
@HPrinz I came across this during the maintenance shift, do we still want to keep it or close it?

anne-ds:
Will close this now as norms.spec.ts is no issue anymore.

PR #2572, opened 2025-02-18
Delete unused functionality

reckseba:
RISDEV-0000

Ran into this while copying over that functionality to ris-adm-vwv. Can this be removed?


Nadav-B:
@reckseba yes it can be removed, thanks for the info

PR #2574, opened 2025-02-18
Display duplicates in status Publishing

kiwikern:
RISDEV-5825


anne-ds (approved):
Nice, that you fixed that and added even a test 🥇

PR #2577, opened 2025-02-18
Remove playwright config complexity

kiwikern:
RISDEV-6806


leonie-koch (commented):
Great, that's so much better! I am wondering if we can get rid of the shared folder completely (I thought there are still some tests inside the shared folder, but it seemed to be empty all the time? 🤯)

kiwikern:
It only had one auth test and the auth setup, but I moved them to `/caselaw`. We don't really need the caselaw folder at all now, but I was a bit afraid that some testreports etc. might break if the path changes.

PR #2587, opened 2025-02-20
Delete source reference

HPrinz:
RISDEV-5898

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/transformer/DecisionTransformer.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/transformer/DecisionTransformer.java
@@ -194,7 +194,26 @@ public class DecisionTransformer {
     addCaselawReferences(updatedDomainObject, builder, currentDto);
     addLiteratureReferences(updatedDomainObject, builder, currentDto);
 
-    return builder.build();
+    // if the source reference has been deleted, remove the link to the source
+    var docUnit = builder.build();
+    if (docUnit.getSource() != null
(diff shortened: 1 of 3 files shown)


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/transformer/DecisionTransformer.java
```diff
@@ -198,23 +198,48 @@ public static DecisionDTO transformToDTO(
...
-    if (coreData.source() != null) {
+      DecisionDTO currentDto, DecisionDTOBuilder<?, ?> builder, DocumentationUnit decision) {
+    if (decision.coreData().source() != null) {
       List<SourceDTO> existingSources =
           currentDto.getSource() != null
               ? new ArrayList<>(currentDto.getSource())
               : new ArrayList<>();
-      Integer rank = existingSources.size() + 1;
 
+      // if the source reference is not included in reference lists (= has been deleted), remove the
```
leonie-koch, line 208: Just for the sake of readability of this condition, can we rewrite it like this?

```
// Check if the first existing source has a reference and remove if it has been deleted in domain object
if (!existingSources.isEmpty()) {
      SourceDTO firstSource = existingSources.getFirst();
      var reference = firstSource.getReference();

      if (reference != null && !decisionContainsReferenceWithId(decision, reference.getId())) {
        firstSource.setReference(null); // Otherwise the source can not be deleted
      }
}
```

HPrinz: Ja, das ist viel besser - danke!

PR #2649, opened 2025-03-18
Ignore duplicates when jDV dup status is set

kiwikern:
RISDEV-6827
Also fixes a bug when GET-requesting a doc unit. The order was wrong so that the duplicates would only be included on the second request.


backend/src/test/java/de/bund/digitalservice/ris/caselaw/integration/tests/DuplicateCheckIntegrationTest.java
```diff
@@ -637,6 +716,53 @@ void deleteDocUnit_withDuplicateRelation_shouldDeleteDuplicateRelation()
           .isEmpty();
     }
 
+    @Test
+    void checkDuplicates_whenRequestingDocUnit_shouldIncludeNewWarningsInResponse() {
```
anne-ds, line 720: @kiwikern Why does this combination create a duplicate? It is only date + filenumber (without doctype)?

anne-ds: I see, it was an older version ✅

PR #2680, opened 2025-03-31
ignored words in text check

Nadav-B:


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/IgnoredTextCheckWordDTO.java
```diff
@@ -0,0 +1,53 @@
...
+  @ManyToOne
+  @JoinColumn(name = "documentation_unit_id")
+  private DocumentationUnitDTO documentationUnit;
+
+  @Column
+  @Size(max = 255)
+  private String word;
+
+  @Column(name = "juris_id")
+  @ToString.Include
```
Nadav-B, line 47: TODO: Remove toString

PR #2683, opened 2025-04-01
[DO NOT MERGE] Showcase V8 line coverage breaks FieldsOfLaw.vue

reckseba:
This PR ships unit tests for `FieldsOfLaw.vue`.

Why? You might ask. According to [Sonar](https://sonarcloud.io/component_measures?metric=coverage&selected=digitalservicebund_ris-frontend%3Asrc%2Fcomponents%2Ffield-of-law%2FFieldsOfLaw.vue&id=digitalservicebund_ris-frontend) the coverage for that file seems rather sufficient.

Yes but. 44% could be improved and the coverage is a false friend as it's only coming via `documentUnitCategories.spec.ts` <- `DocumentUnitCategories.vue` <- `DocumentUnitContentRelatedIndexing.vue` where `FieldsOfLaw.vue` is imported but not tested.

Therefore I added stubbing to get a clear view on the coverage.

According the changes in `frontend/vite.config.ts`:
- I added the `text` reporter only to view the coverage in the console (remove before merging)
- I needed to add `reportOnFailure` bc:
  - There is one unrelated unit test failing and
  - one within the domain ('Shows norms when required') which might point to a bug in the Tree (not further discussed).

What we can observe here is, that when running `npm run coverage` the coverage _drops_ to: ` FieldsOfLaw.vue                 |   16.93 |     92.3 |      75 |   16.93 | 1-182` with a very weird line coverage that is more than suspicious.

You can run `npx vite preview --outDir coverage/lcov-report` and open `http://localhost:3000` to get a html rendering (Just adding for completeness of the description - I am sure you are familiar).

Here we are. That is the problem. I have no idea so far how to fix it. Do you?

Interestingly I have the following observations to share:

When you run `npm i -D @vitest/coverage-istanbul@3.0.9` and change the `coverage.provider` to `istanbul` in the `vite.config.ts` the problem is fixed (After deleting `LinkElement.vue` which I [raised](https://digitalservicebund.slack.com/archives/C04C8MEGUBX/p1743152499765989) some time ago). But I am not suggesting a change as this would be a major shift. Just an observation.

Upgradig all vitest packages to 3.1.1 did not solve it.

To make it even more weird: Set row 39 to `describe.skip("FieldsOfLaw", () => {` and watch that coverage is up to incredibly 100%.

### DIFF
--- a/frontend/test/components/documentUnit/documentUnitCategories.spec.ts
+++ b/frontend/test/components/documentUnit/documentUnitCategories.spec.ts
@@ -25,6 +25,9 @@ function renderComponent() {
       },
       global: {
         directives: { "ctrl-enter": onSearchShortcutDirective },
+        stubs: {
+          FieldsOfLaw: true,
+        },
         plugins: [
           createTestingPinia({
             initialState: {
--- a/frontend/test/components/fieldOfLaw/FieldsOfLaw.spec.ts
+++ b/frontend/test/components/fieldOfLaw/FieldsOfLaw.spec.ts
@@ -1,0 +1,529 @@
+import { createTestingPinia } from "@pinia/testing"
+import { userEvent } from "@testing-library/user-event"
+import { render, screen, waitFor } from "@testing-library/vue"
+import FieldsOfLawVue from "@/components/field-of-law/FieldsOfLaw.vue"
+import DocumentUnit from "@/domain/documentUnit"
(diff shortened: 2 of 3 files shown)


Nadav-B:
@reckseba  thanks for you finding. In the scope I did the following:

1. Importing vwv field of low refactoring to caselaw
2. Opened a PR for istnabul test coverage 

I will close the PR for now

PR #2685, opened 2025-04-01
Retry docunit deletion in e2e fixture

kiwikern:
RISDEV-7275
When a doc unit is deleted its associated duplicate relations need to be deleted as well. If two related doc units are deleted at the same time, the second one will try to delete a stale reference that was already deleted as we are using optimistic locking.

### DIFF
--- a/frontend/test/e2e/caselaw/fixtures.ts
+++ b/frontend/test/e2e/caselaw/fixtures.ts
@@ -1,4 +1,10 @@
-import { Locator, Page, test } from "@playwright/test"
+import {
+  APIRequestContext,
+  Cookie,
+  Locator,
+  Page,
+  test,
+} from "@playwright/test"
 import dayjs from "dayjs"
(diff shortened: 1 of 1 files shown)


frontend/test/e2e/caselaw/fixtures.ts
```diff
@@ -30,28 +36,54 @@ type MyFixtures = {
...
+    },
+  )
 
-    await use(documentNumber)
-    const deleteResponse = await request.delete(
+  if (!deleteResponse.ok()) {
+    console.log(
+      `Deletion for ${documentNumber} failed with status ${deleteResponse.status()}, retrying deletion...`,
+    )
+    const retryWaitDuration = Math.floor(Math.random() * 1900) + 100
```
HPrinz, line 59: Super Umsetzung @kiwikern! Für diese magic numbers fänd ich noch einen Kommentar toll, z.B. `retry at a random time between 0.1 and 2 seconds`

PR #2695, opened 2025-04-04
Ignore and add word to doc unit

Nadav-B:


backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/TextCheckService.java
```diff
@@ -25,28 +27,22 @@
...
-    this.documentationUnitService = documentationUnitService;
-    this.documentationOfficeService = documentationOfficeService;
+
+    this.documentationUnitRepository = documentationUnitRepository;
     this.ignoredTextCheckWordRepository = ignoredTextCheckWordRepository;
   }
 
   public List<Match> check(String text) {
-    return requestTool(text);
+    return requestTool(text).stream().toList();
```
HPrinz, line 45: Is this stream + list intentional @Nadav-B ?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/TextCheckService.java
```diff
@@ -55,119 +51,81 @@ protected List<Match> requestTool(String text) {
 
   public List<Match> checkWholeDocumentationUnit(UUID id)
       throws DocumentationUnitNotExistsException {
+
     List<Match> allMatches = new ArrayList<>();
```
HPrinz, line 55: I suggest moving the initialization into the if-clause

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/TextCheckService.java
```diff
@@ -55,119 +51,81 @@ protected List<Match> requestTool(String text) {
...
 
-    Documentable documentable = documentationUnitService.getByUuid(id);
+    Documentable documentable = documentationUnitRepository.findByUuid(id);
 
-    if (!(documentable instanceof DocumentationUnit documentationUnit)) {
-      throw new UnsupportedOperationException(
-          "Check not supported for Documentable type: " + documentable.getClass());
-    }
```
HPrinz, line 65: Did you intend to remove the Exception description (`Check not supported for Documentable type: " + documentable.getClass()`) ?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/TextCheckService.java
```diff
@@ -235,30 +184,55 @@ protected TextCheckCategoryResponse checkCategoryByHTML(
...
+        ignoredTextCheckWordRepository.findByDocumentationUnitIdOrByGlobalWords(
+            words, documentationUnitId);
+
+    Map<String, List<IgnoredTextCheckWord>> groupedByWord =
+        globalAndDocumentationUnitIgnoredWords.stream()
+            .collect(Collectors.groupingBy(IgnoredTextCheckWord::word));
+
+    if (documentationUnitId == null) {
+      return matches;
+    }
```
HPrinz, line 228: Can't we do this at the method start already?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/TextCheckService.java
```diff
@@ -235,30 +184,55 @@ protected TextCheckCategoryResponse checkCategoryByHTML(
...
+    return matches.stream()
+        .map(
+            match -> {
+              List<IgnoredTextCheckWord> ignoredWords =
+                  groupedByWord.getOrDefault(match.word(), null);
+              return match.toBuilder().ignoredTextCheckWords(ignoredWords).build();
```
HPrinz, line 234: I'm thinking: Does it really make sense to send a **list** of ignoredWords for each match? Maybe we can calculate a "winning" ignoreWord (which would probably be the global) and simply remove the other?
Why I stumbled upon this: If we handle a list, we need to handle and render cases where a match is ignored by both jdv/global and decision and I don't see why this would make sense functionally. @Nadav-B - what do you think?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/textcheck/ignored_words/IgnoredTextCheckWordRequest.java
```diff
@@ -0,0 +1,3 @@
+package de.bund.digitalservice.ris.caselaw.domain.textcheck.ignored_words;
+
+public record IgnoredTextCheckWordRequest(String word) {}
```
HPrinz, line 3: Can't we simply send a String word then? 🤔

Nadav-B: I agree, however thought behind that it might be handy later on if we add the ignored type level for the word.

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/TextCheckService.java
Nadav-B: No thanks for pointing it out. I removed it while refactoring. I added in test the expected message.

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/TextCheckService.java
HPrinz: Yes, the client needs to know why the word is ignored (currently global/jdv or documentationUnit). If I understand the current implementation correctly, we can give the client multiple reasons. Currently we could express that the word is ignored by both jDV and documentationUnit. 
I'm still not sure if the client can meaningfully use these two reasons (e.g. display two modals?). But maybe I missed something. Maybe let's sync about it

PR #2860, opened 2025-06-18
Update node to 22.16.0

anne-ds:
RISDEV-0000


anne-ds:
Was just a test if the incompatibilty with msw is resolved now.

PR #2868, opened 2025-06-19
Fix doc unit store domain class instantiation

kiwikern:
RISDEV-8322


anne-ds (approved):
It think it looks much better. Thanks. 🚀

PR #2873, opened 2025-06-19
Refactor integration test setup

kiwikern:
RISDEV-8290


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/eurlex/EurLexSOAPSearchService.java
```diff
@@ -89,7 +90,7 @@ public Page<SearchResult> getSearchResults(
     if (!documentationOffice.abbreviation().equals("DS")
         && !documentationOffice.abbreviation().equals("BGH")
         && !documentationOffice.abbreviation().equals("BFH")) {
-      return Page.empty();
+      return new PageImpl<>(List.of(), PageRequest.of(0, PAGE_SIZE), 0);
```
kiwikern, line 93: This leads to a `HttpMessageNotWritableException` in the integration test after removing the mocking. Does this work in the real application?

PR #3061, opened 2025-08-20
Add process steps information to search result

Nadav-B:


backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/DocumentationUnitListItem.java
```diff
@@ -24,6 +25,8 @@
  * @param hasAttachments Search result has attachments
  * @param hasHeadnoteOrPrinciple Search result has headnote or principle
  * @param source the source the documentation unit was created from
+ * @param processSteps (Prozessschritte) history of the search result
+ * @param currentProcessStep (Prozessschritt) of the search result
```
HPrinz, line 29: Maybe "search result" -> "process steps"?

PR #3096, opened 2025-08-26
Current and previous process steps in result list

Nadav-B:
RISDEV-8982

* Flatten the current and last component for process step only
* Don't load the current process step on result list

### DIFF
--- a/frontend/src/components/DocumentUnitInfoPanel.vue
+++ b/frontend/src/components/DocumentUnitInfoPanel.vue
@@ -130,7 +130,10 @@ watchEffect(() => {
         :previous-process-step="props.documentUnit.previousProcessStep"
       />
       <IconBadge
-        v-if="props.documentUnit.currentProcessStep && processStepsEnabled"
+        v-if="
+          props.documentUnit.currentDocumentationUnitProcessStep &&
+          processStepsEnabled
+        "
         background-color="bg-white"
         border-color="border-gray-800"
         class="px-8"


frontend/src/components/DocumentUnitInfoPanel.vue
```diff
@@ -121,9 +121,13 @@ watchEffect(() => {
...
+      <CurrentAndPreviousProcessStepBadge
         v-if="processStepsEnabled"
-        :process-steps="props.documentUnit.processSteps"
+        :current-process-step="
+          props.documentUnit.currentDocumentationUnitProcessStep?.processStep
+        "
+        :previous-process-step="props.documentUnit.previousProcessStep"
       />
       <IconBadge
         v-if="
```
HPrinz, line 133: In Line 134: I think it should be `props.documentUnit.currentDocumentationUnitProcessStep?.user`, right?

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/DocumentationUnitListItemDTO.java
```diff
@@ -49,7 +49,5 @@ public interface DocumentationUnitListItemDTO {
 
   List<SourceDTO> getSource();
 
-  DocumentationUnitProcessStepDTO getCurrentProcessStep();
```
HPrinz, line 52: I'm thinking: I know this has been removed to save reading the first step from the DB multiple times. But maybe we should go back to using the currentProcessStep DTO for consistency reasons. It's the same field that is used for the search filter.

frontend/src/components/DocumentUnitInfoPanel.vue
Nadav-B: @HPrinz dear Hanna, the CurrentAndPreviousProcessStepBadge only needs the previous and current process steps,  I flatten it for reuse and testing reasons.

The user is passed in the [next line](https://github.com/digitalservicebund/ris-backend-service/blob/a4c441c0d87e880827bee1177eb8da5e66ae3449/frontend/src/components/DocumentUnitInfoPanel.vue#L132-L132)

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/DocumentationUnitListItemDTO.java
Nadav-B: Thinking out loud: overall I find the DocumentationUnitListItemDTO interface being the root interface for DecisionDto, DocumentationUnitDto, and PendingProceedingDto misleading, since it’s really just another extension of the root case. As I don't need it in our result list, I didn't want to load the object with jpa. 

Didn’t we once say we were going to change the `previous` and the `current` and then drop the processSteps list so it’s only called when needed? Maybe that’s something we can do properly later, what do you think?

PR #3103, opened 2025-08-27
Fix category import bug

kiwikern:
RISDEV-9003
The category import of editable lists in content related indexing did not remove the ids correctly. -> Instead of new copies being created the existing objects were moved to a different doc unit.


anne-ds (commented):
Looks good to me. Great that you found the reason and added so many tests! 🎖️ 👏

PR #3105, opened 2025-08-27
Sentry user exception

Nadav-B:


backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/DocumentationUnitService.java
```diff
@@ -478,9 +477,10 @@ public DocumentationUnit getByDocumentNumberWithUser(String documentNumber, Oidc
   private Slice<DocumentationUnitListItem> retrieveCurrentProcessStepUser(
       Slice<DocumentationUnitListItem> documentationUnitListItems, OidcUser oidcUser) {
 
-    Map<UUID, User> userIdMap =
-        userService.getUsers(oidcUser).stream()
-            .collect(Collectors.toMap(User::id, Function.identity()));
+    Map<UUID, User> userIdMap = new HashMap<>();
+    for (User user : userService.getUsers(oidcUser)) {
```
rvp-c, line 481: I find the name of the of the method not good. we have now a method which returns a list of users for one user? I think the method returns other user of the same doc office as the user? My idea is getUsersAtTheSameDocOfficeAs(oidcUser)?

rvp-c: Second one? With this solution you always override the value in the map, right? So the last entry wins.

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/User.java
```diff
@@ -11,4 +12,11 @@ public record User(
     String email,
     DocumentationOffice documentationOffice,
     List<String> roles,
-    String initials) {}
+    String initials) {
+
+  @NotNull
+  @Override
+  public String toString() {
+    return "User[id=%s]".formatted(id);
```
rvp-c, line 20: Great. Doesn't know this is possible!

backend/src/main/java/de/bund/digitalservice/ris/caselaw/domain/User.java
```diff
@@ -11,4 +12,11 @@ public record User(
     String email,
     DocumentationOffice documentationOffice,
     List<String> roles,
-    String initials) {}
+    String initials) {
+
+  @NotNull
```
rvp-c, line 17: Why do we need this annotation? And why it is a jetbrain annotation? I think with the implementation it can't be null.

rvp-c (commented):
Only points to think about.

PR #3107, opened 2025-08-27
Select the longest segment user group

Nadav-B:
RISDEV-9053

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/KeycloakUserService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/KeycloakUserService.java
@@ -8,6 +8,7 @@ import de.bund.digitalservice.ris.caselaw.domain.UserGroup;
 import de.bund.digitalservice.ris.caselaw.domain.UserGroupService;
 import de.bund.digitalservice.ris.caselaw.domain.UserService;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -78,9 +79,11 @@ public class KeycloakUserService implements UserService {
   public Optional<UserGroup> getUserGroup(OidcUser oidcUser) {
     List<String> userGroups = Objects.requireNonNull(oidcUser.getAttribute("groups"));
     var matchingUserGroup =
-        this.userGroupService.getAllUserGroups().stream()
+        Optional.ofNullable(this.userGroupService.getAllUserGroups())
+            .orElse(Collections.emptyList())
+            .stream()
             .filter(group -> userGroups.contains(group.userGroupPathName()))
-            .findFirst();
+            .max(Comparator.comparingInt(group -> group.userGroupPathName().split("/").length));
     if (matchingUserGroup.isEmpty()) {
       LOGGER.warn(
           "No doc office user group associated with given Keycloak user groups: {}", userGroups);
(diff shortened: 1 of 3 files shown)


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/KeycloakUserService.java
```diff
@@ -78,9 +79,11 @@ public List<User> getUsers(OidcUser oidcUser) {
   public Optional<UserGroup> getUserGroup(OidcUser oidcUser) {
     List<String> userGroups = Objects.requireNonNull(oidcUser.getAttribute("groups"));
     var matchingUserGroup =
-        this.userGroupService.getAllUserGroups().stream()
+        Optional.ofNullable(this.userGroupService.getAllUserGroups())
+            .orElse(Collections.emptyList())
+            .stream()
             .filter(group -> userGroups.contains(group.userGroupPathName()))
-            .findFirst();
+            .max(Comparator.comparingInt(group -> group.userGroupPathName().split("/").length));
```
Nadav-B, line 86: @kiwikern  What to do if non of them have a "/" in the path? Shall we pick up the first one in this case?

In the exmaple there were also groups like: `intrenal`

Nadav-B: @kiwikern i am closing this pull request and openning a new one, as the longest segement is redundant to the mapping.

@kiwikern / @anne-ds Here we have [getUserGroup](https://github.com/digitalservicebund/ris-backend-service/blob/8308af63030d832773982e60a62e3c79c4435d55/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/KeycloakUserService.java?utm_source=chatgpt.com#L78). However, there is also [getDocumentationOfficeConfigUserGroups](https://github.com/digitalservicebund/ris-backend-service/blob/8308af63030d832773982e60a62e3c79c4435d55/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/BareIdUserApiService.java?utm_source=chatgpt.com#L205). Is there any reason, why in keycloak service, we don't select by the mapping, and instead just doing findFirst?  I would like to have one mehthod that is in charge of the group selection and remove the redundant if so in bare user api.

kiwikern: We have the userGroup config that maps the doc office string (e.g. "/caselaw/BGH/Intern" or "/caselaw/BGH") to a doc office group entity/id. (1<->n)

We have users that have have 0..n groups associated with them ("/caselaw/BGH" and "/caselaw/BGH/Intern").

Current approaches:
a) KeycloakService: Find first userGroupConfig that is equal to one of the user's bareId groups. If none, throw.
b) BareIdService: Find all matches between user groups and config. If count!=1, throw.

So in my understanding, both services map to the userGroup. For a quick win, I still think the longest segment selection would be better than `findFirst` (KeyCloak) but more robust than the ==1 restriction against misconfiguration (BareId).

And more effort would be to rethink and maybe refactor the whole UserGroup config setup, but I don't think we have time for that.

Nadav-B: Thanks for the info @kiwikern . I made some [changes](https://github.com/digitalservicebund/ris-backend-service/pull/3152?utm_source=chatgpt.com) only to the user group, extracting this part so we can reusing what was there. 

Info: Given `/caselaw/BGH/Intern` the user api will fetch both BGH and Intern. So the longest segment would also work here. 

I will leave both be for the moment, till we consolidate this.

PR #3133, opened 2025-09-01
Prohibit deletion of published doc units

kiwikern:
RISDEV-8456
If a published doc unit were to be deleted, it would not be possible to withdraw it from the portal.


anne-ds:
@kiwikern Nice implementation 🎖️. What is the additional logging for?

kiwikern:
@anne-ds I would like to have logs for every business-critical operation. If we a doc unit is deleted, we also lose the historylogs (also: I don't think we should base our debugging processes on business logic/entities). If we want to investigate a doc unit, I want to be able to search for the id / number and find a lot of corresponding logs to it to understand what has happend with it.

PR #3158, opened 2025-09-05
Write failure history log despite exception

elaydis:
RISDEV-8460
RISDEV-8456


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/PortalPublicationService.java
```diff
@@ -76,7 +76,12 @@ public PortalPublicationService(
    * @throws PublishException if the LDML could not be saved in the bucket
    * @throws ChangelogException if the changelog cannot be generated or saved
    */
-  @Transactional
+  @Transactional(
```
anne-ds, line 79: @elaydis I initially added the annotation to solve another issue (saving management data). So the complete removal of this annotation solved your issue. And I found a much better solution for my original issue. 
Root cause of my issue: The builder pattern that doesn't create a deep copy of the object (hence properties where missing afterwards, in my case managementData).
Why I invested into another solution: The "noRollbackFor" is quite dangerous in my opinion because it can create inconsistent data where the root cause is almost impossible to find...

PR #3169, opened 2025-09-09
Update LDML to latest version 1.0 (2018)

anne-ds:
RISDEV-9114

Don't merge before synching with Portal.


anne-ds:
Moved to ris-data-migration ✅

PR #3179, opened 2025-09-12
[RISDEV-9094] Fix error count

duskox:
- fixed error count when some error words are ignored
- refactor for readability

### DIFF
--- a/frontend/src/components/text-check/HandoverTextCheckView.vue
+++ b/frontend/src/components/text-check/HandoverTextCheckView.vue
@@ -89,10 +89,13 @@ onBeforeMount(async () => {
                 <dl class="my-16">
                   <div
                     class="grid grid-cols-[auto_1fr] gap-x-16 px-0"
-                    data-testid="total-text-check-errors"
+                    data-testid="total-text-check-errors-container"
                   >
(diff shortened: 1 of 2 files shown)


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/transformer/TextCheckResponseTransformer.java
```diff
@@ -122,4 +119,23 @@ public static TextCheckResponse transformToDomain(
 
     return matches;
   }
+
+  private static String extractWordFromMatchRecord(
```
Nadav-B, line 123: Maybe rename to : getMatchWord?

We could now call the same method in: Line `86`

Nadav-B:
@duskox this is our [link](https://github.com/digitalservicebund/ris-backend-service/blob/45ebe2bd02b55a4024bdfc90b3b0d420a579da6e/frontend/test/e2e/caselaw/decision/handover/handover-documentation-unit.spec.ts#L507-L508) to related e2e test:

duskox:
@Nadav-B I see that `e2e-tests` pipeline is skipped?

PR #3190, opened 2025-09-17
Clear process steps when assigning new docoffice

leonie-koch:
RISDEV-9167
This logic applies when a document's Documentation Office is reassigned.

scenario 1: docoffice A assinging to docoffice B -> all previous process steps are cleared. This action deletes all underlying documentationUnitProcessStep data. For our history logs the loss of the documentationUnitProcessSteps is acceptable, because we only need these documentationUnitProcessSteps to retrieve the user data dynamically. As the process step change the history logs save a description like 'Schritt geändert: [from] -> [to]', which is hardcoded information in the description field and will therefore be there even though the underlying information is not there anymore. Only the user data can not be retrieved, but in this case it is fine, because the docoffice B is not allowed to see user information from docoffice A anyways, they will only see 'Person geändert'.

scenario 2: docoffice B assigns docunit back to docoffice A -> when assigning BACK to docoffice A, this information can not be restored, because the underlying documentationUnitProcessStep data is gone. Meaning, that we still do only show 'Person geändert', even though the user from docoffice A should be able to see the user information. We took this tradeoff, because this is a very rare edge case and to properly handle previous process steps for other docoffices after reassigning, we need to store the docoffice information seperately in the documentationUnitProcessStep and not in the user of the documentationUnitProcessStep. (Because the user will not always be there to provide that information).

This trade-off was made because this 'reassign back' scenario is a very rare edge case. The proper solution would require storing the documentation office information separately within the DocumentationUnitProcessStep entity to preserve history, which was deemed too much effort for this very edge case.

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/PostgresDocumentationUnitRepositoryImpl.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/PostgresDocumentationUnitRepositoryImpl.java
@@ -565,6 +565,8 @@ public class PostgresDocumentationUnitRepositoryImpl implements DocumentationUni
 
     DocumentationUnitProcessStep currentDocunitProcessStepFromFrontend =
         documentationUnit.currentDocumentationUnitProcessStep();
+    List<DocumentationUnitProcessStep> docunitProcessStepsFromFrontend =
+        documentationUnit.processSteps();
     DocumentationUnitProcessStepDTO currentDocumentationUnitProcessStepDTOFromDB =
         documentationUnitDTO.getCurrentProcessStep();
 
@@ -593,7 +595,10 @@ public class PostgresDocumentationUnitRepositoryImpl implements DocumentationUni
       processStepChanged = true;
       DocumentationUnitProcessStepDTO newDocumentationUnitProcessStepDTO =
           createAndSaveNewProcessStep(
-              documentationUnitDTO, processStepDTO, currentDocunitProcessStepFromFrontend);
+              documentationUnitDTO,
+              processStepDTO,
+              currentDocunitProcessStepFromFrontend,
+              docunitProcessStepsFromFrontend);
 
       if (stepChanged) {
         String description =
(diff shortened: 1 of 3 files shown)


backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/PostgresDocumentationUnitRepositoryImpl.java
```diff
@@ -683,7 +688,18 @@ private boolean userChanged(
...
       ProcessStepDTO processStepDTO,
-      DocumentationUnitProcessStep currentDocunitProcessStepFromFrontend) {
+      DocumentationUnitProcessStep currentDocunitProcessStepFromFrontend,
+      List<DocumentationUnitProcessStep> docunitProcessStepsFromFrontend) {
+
+    // when assigning a new doc office, all previous process steps are removed
+    boolean isOwnerChangeScenario =
+        (docunitProcessStepsFromFrontend != null
+            && docunitProcessStepsFromFrontend.isEmpty()
+            && "Neu".equals(processStepDTO.getName()));
```
Nadav-B, line 698: 1. Could be extract for a static method, good for direct unit testing later
2. If we cahnge the name of the process step this logic would silently fail. I thought maybe we use the constnat process step id we set on flyway script for mapping. As we use it in both doc service and the process step maybe we could extract it to a static variable and reuse.

backend/src/main/java/de/bund/digitalservice/ris/caselaw/adapter/database/jpa/PostgresDocumentationUnitRepositoryImpl.java
```diff
@@ -683,7 +688,18 @@ private boolean userChanged(
...
+
+    if (isOwnerChangeScenario) {
+      documentationUnitDTO.getProcessSteps().clear();
```
Nadav-B, line 701: Validation for list exists not null, if it's empty we will get an error here :)

leonie-koch: good catch!!

PR #3207, opened 2025-09-22
Fix code smells

leonie-koch:
RISDEV-0000


leonie-koch:
Closing this PR, as it is outdated

PR #3213, opened 2025-09-23
[RISDEV-9180] Error categories show same red wawy underline

duskox:
RISDEV-9180

### DIFF
--- a/frontend/src/styles/text-check.scss
+++ b/frontend/src/styles/text-check.scss
@@ -1,28 +1,28 @@
 text-check {
-  text-decoration: underline wavy #CD5038 2px;
+  text-decoration: underline wavy #cd5038 2px;
   transition: background-color 0.25s ease-in-out;
 
   &:hover {
-    background: rgba($color: #F5DCD7, $alpha: 20%);
+    background: rgba($color: #f5dcd7, $alpha: 20%);
   }
 
   &[type="style"],
   &[type="grammar"],
   &[type="typographical"],
   &[type="misspelling"] {
-    text-decoration-color: #CD5038;
+    text-decoration-color: #cd5038;
 
     &:hover {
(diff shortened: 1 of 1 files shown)


frontend/src/styles/text-check.scss
```diff
@@ -1,41 +1,28 @@
...
-  &[type="style"] {
-    border-bottom: 2px solid #9d8eff;
-
-    &:hover {
-      background: rgba($color: #9d8eff, $alpha: 20%) !important;
-    }
+    background: rgba($color: #f5dcd7, $alpha: 20%);
   }
 
+  &[type="style"],
```
Nadav-B, line 9: @duskox If we only have one red color and the type category is no longer important, we can remove those types `  &[type="style"],` as its redunant:  in favour of the parent one line 2

Nadav-B:
@duskox Please adjust also here at https://github.com/digitalservicebund/ris-backend-service/blob/064e935074e5b106b4137215c06ea301b9b59a0d/frontend/test/e2e/caselaw/text-check/text-check-category.spec.ts#L13-L13 as we test also by different colors.

frontend/src/styles/text-check.scss
```diff
@@ -1,41 +1,28 @@
 text-check {
-  border-bottom: 2px solid #e86a69;
-  transition: 0.25s ease-in-out;
+  text-decoration: underline wavy #cd5038 2px;
```
Nadav-B, line 2: @duskox seems like we need to adjust the size of the decoration line, as it is too thick : https://www.figma.com/design/9CiLT0uhSOQgs257h4DW8q/EA-13---14-Caselaw?node-id=3268-69618&t=P5uZMsxzbPlnhaGe-0

duskox: You think it is too thick? I had it at 1px and then it didn't seem like the line in Figma? That's why I made it 2px.

PR #3250, opened 2025-10-01
Fix code smell

duskox:
Sonarcube reposts a code smell where
we should use `for..of` instead of `forEach`
RISDEV-0000


duskox:
Closed because we prefer `forEach` and sonarcube rule should be ignored.

